PR #30, opened 2025-01-13
feat: revamp validator support for catalog 

abhimanyubabbar:
## Description of the change

Validation support for the catalog needs to be revamped as we need more fine grained control over the validation logic and also to apply validations at different layers.

The existing validations were clubbing all the entities together. Rather than working at the entity level, it was working at the whole catalog level which meant it was breaking the SRP principle. This made testing very hard.

The idea of the change is to introduce `validation rules` at entity level and combine them to perform validation at the catalog level. This makes the code testable as it has only one reason to change. In addition to this, the new way would be more maintainable as number of entities in the catalog grows and would require custom validation logic.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/cli/pkg/validate/entity/testutils/catalog_factory.go
+++ b/cli/pkg/validate/entity/testutils/catalog_factory.go
@@ -1,0 +1,41 @@
+package testutils
+
+import "github.com/rudderlabs/rudder-iac/cli/pkg/localcatalog"
+
+type DataCatalogFactory struct {
+	dc *localcatalog.DataCatalog
+}
+
+func NewDataCatalogFactory() *DataCatalogFactory {
+	return &DataCatalogFactory{
+		dc: &localcatalog.DataCatalog{},
+	}
+}
+
+func (f *DataCatalogFactory) WithEvent(group string, event *localcatalog.Event) *DataCatalogFactory {
+	if f.dc.Events == nil {
+		f.dc.Events = make(map[localcatalog.EntityGroup][]*localcatalog.Event)
+	}
+	f.dc.Events[localcatalog.EntityGroup(group)] = append(f.dc.Events["default"], event)
+	return f
+}
+
+func (f *DataCatalogFactory) WithProperty(group string, prop *localcatalog.Property) *DataCatalogFactory {
+	if f.dc.Properties == nil {
+		f.dc.Properties = make(map[localcatalog.EntityGroup][]*localcatalog.Property)
+	}
+	f.dc.Properties[localcatalog.EntityGroup(group)] = append(f.dc.Properties["default"], prop)
+	return f
+}
+
+func (f *DataCatalogFactory) WithTrackingPlan(group string, tp *localcatalog.TrackingPlan) *DataCatalogFactory {
+	if f.dc.TrackingPlans == nil {
+		f.dc.TrackingPlans = make(map[localcatalog.EntityGroup]*localcatalog.TrackingPlan)
+	}
+	f.dc.TrackingPlans[localcatalog.EntityGroup(group)] = tp
+	return f
+}
+
+func (f *DataCatalogFactory) Build() *localcatalog.DataCatalog {
+	return f.dc
+}
--- a/cli/pkg/validate/entity/testutils/trackingplan_factory.go
+++ b/cli/pkg/validate/entity/testutils/trackingplan_factory.go
@@ -1,0 +1,42 @@
+package testutils
+
+import "github.com/rudderlabs/rudder-iac/cli/pkg/localcatalog"
+
+type LocalCatalogTrackingPlanFactory struct {
+	tp *localcatalog.TrackingPlan
+}
+
+func NewLocalCatalogTrackingPlanFactory() *LocalCatalogTrackingPlanFactory {
+	return &LocalCatalogTrackingPlanFactory{
+		tp: &localcatalog.TrackingPlan{},
+	}
+}
+
+func (f *LocalCatalogTrackingPlanFactory) WithLocalID(localID string) *LocalCatalogTrackingPlanFactory {
+	f.tp.LocalID = localID
+	return f
+}
+
+func (f *LocalCatalogTrackingPlanFactory) WithName(name string) *LocalCatalogTrackingPlanFactory {
+	f.tp.Name = name
+	return f
+}
+
+func (f *LocalCatalogTrackingPlanFactory) WithDescription(description string) *LocalCatalogTrackingPlanFactory {
+	f.tp.Description = description
+	return f
+}
+
+func (f *LocalCatalogTrackingPlanFactory) WithRule(rule *localcatalog.TPRule) *LocalCatalogTrackingPlanFactory {
+	f.tp.Rules = append(f.tp.Rules, rule)
+	return f
+}
+
+func (f *LocalCatalogTrackingPlanFactory) WithExpandedEvent(eventProp *localcatalog.TPEvent) *LocalCatalogTrackingPlanFactory {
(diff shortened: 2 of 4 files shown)


cli/internal/cmd/trackingplan/apply/apply.go
```diff
@@ -39,21 +42,25 @@ func NewCmdTPApply() *cobra.Command {
 			$ rudder-cli tp apply --loc </path/to/dir or file> --dry-run
 		`),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			fmt.Println()
```
fxenik, line 45: Is this just for formatting?

cli/internal/cmd/trackingplan/apply/apply.go
```diff
@@ -39,21 +42,25 @@ func NewCmdTPApply() *cobra.Command {
...
 				return fmt.Errorf("reading catalog failed in pre-step: %w", err)
 			}
 
-			if err := validate.ValidateCatalog(validate.DefaultValidators(), localcatalog); err != nil {
-				return fmt.Errorf("validating catalog: %w", err)
+			if errs := validate.NewCatalogValidator().Validate(localcatalog); errs != nil {
+				fmt.Println("catalog errors:")
+				fmt.Println(errs.Error())
+
+				return errors.New(emoji.Sprintf("catalog invalid :x:"))
```
fxenik, line 57: nit: let's move emojis at the start of the line. Personally, I would also consider not adding the emoji in the error itself, and use it as part of the validation report that's printed before. The reason is that this error is handled in a generic way by the framework and already prints an "Error" label.

cli/internal/cmd/trackingplan/apply/apply.go
```diff
@@ -39,21 +42,25 @@ func NewCmdTPApply() *cobra.Command {
...
 			if err != nil {
 				return fmt.Errorf("reading catalog failed in pre-step: %w", err)
 			}
 
-			if err := validate.ValidateCatalog(validate.DefaultValidators(), localcatalog); err != nil {
-				return fmt.Errorf("validating catalog: %w", err)
+			if errs := validate.NewCatalogValidator().Validate(localcatalog); errs != nil {
+				fmt.Println("catalog errors:")
+				fmt.Println(errs.Error())
+
```
fxenik, line 56: It looks like we're reporting errors in both apply and validate commands. It's better if reporting the validation results is moved to a dedicated function (in Validator itself maybe?) and use that in both places.

cli/internal/cmd/trackingplan/validate/validate.go
```diff
@@ -5,23 +5,21 @@ import (
...
-		Value: "validate",
-	})
+	log = logger.New("trackingplan.validate")
 )
 
 func NewCmdTPValidate() *cobra.Command {
 	var catalogDir string
 
-	validators := DefaultValidators()
+	// validators := DefaultValidators()
```
fxenik, line 22: Is this necessary?

cli/internal/cmd/trackingplan/validate/validate.go
```diff
@@ -30,52 +28,29 @@ func NewCmdTPValidate() *cobra.Command {
 			$ rudder-cli tp validate --loc <path-to-catalog-dir or file>
 		`),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			fmt.Println()
```
fxenik, line 31: I'm a lit bit skeptical of adding this Println before the output in selected commands only. If we're making this kind of presentation decisions, it's better if they are handled the same way across commands. Personally, I would prefer to skip it.

cli/pkg/validate/entity/entity_validator.go
```diff
@@ -0,0 +1,67 @@
+package entity
```
fxenik, line 1: One concern with this framework is that there is no clear separation of which part if generic and which part is specific to the Data Catalog. If we're looking to extend the CLI to other resources, they would probably need a similar framework for validations, with many common functionalities (e.g checking for unique ids per type, validate references etc).

This approach looks like it's operating on the DataCatalog struct. I get the feeling this dependency is forced by the loader which might need a similar generic framework for parsing YAMLs into structs while considering a common structure like entity groups with their metadata, entity ids, using references etc.

cli/pkg/validate/entity/entity_validator.go
```diff
@@ -0,0 +1,67 @@
...
+	log = logger.New("entity.validate")
+)
+
+var (
+	// Common Errors
+	ErrDuplicateByID           = errors.New("duplicate entity by id")
+	ErrDuplicateByName         = errors.New("duplicate entity by display_name")
+	ErrDuplicateByNameType     = errors.New("duplicate entity by name and type")
+	ErrMissingRequiredKeysID   = errors.New("missing required key id")
+	ErrMissingRequiredKeysName = errors.New("missing required key display_name")
```
fxenik, line 20: Some of these errors look generic (e.g duplicate entity by id) while others look specific to this type of entities (e.g display_name needs to be unique). We should clearly separate those two and allow specific entities to have their own specific validation rules if necessary.

cli/pkg/validate/entity/event.go
```diff
@@ -0,0 +1,176 @@
...
+var (
+	TrackEventTypes    = []string{"track"}
+	NonTrackEventTypes = []string{"identify", "group", "page", "screen"}
+)
+
+var (
+	_ TypedCatalogEntityValidator[*localcatalog.Event] = &EventValidator{}
+	_ ValidationRule[*localcatalog.Event]              = &EventRequiredKeysRule{}
+	_ ValidationRule[*localcatalog.Event]              = &EventDuplicateKeysRule{}
+)
```
fxenik, line 19: Why do we need these?

cli/pkg/validate/entity/event.go
```diff
@@ -0,0 +1,176 @@
...
+	ev.rules = append(ev.rules, rule)
+}
+
+func (ev *EventValidator) Validate(dc *localcatalog.DataCatalog) (errs []ValidationError) {
+	log.Info("validating event entities in the catalog")
+
+	for group, events := range dc.Events {
+		for _, event := range events {
+
+			reference := fmt.Sprintf(
```
fxenik, line 35: At some point it might be nice to be able to annotate the reference with the source file and line this reference was declared in, to help anyone fixing errors to know exactly where this reference was introduced.

cli/pkg/validate/entity/event.go
```diff
@@ -0,0 +1,176 @@
...
+			for _, rule := range ev.rules {
+				errs = append(errs, rule.Validate(reference, event, dc)...)
+			}
+		}
+	}
+
+	return errs
+}
+
+type EventRequiredKeysRule struct {
```
fxenik, line 50: Given you take a mostly functional approach here, and have all logic in a validation functions (which is great), can we avoid these empty structs? Can we have a function signature instead and pass functions directly?

cli/pkg/validate/entity/property.go
```diff
@@ -0,0 +1,144 @@
...
+	var errors []ValidationError
+
+	for group, properties := range dc.Properties {
+		for _, property := range properties {
+
+			reference := fmt.Sprintf(
+				"#/properties/%s/%s",
+				group,
+				property.LocalID,
+			)
```
fxenik, line 38: Something to consider about references, probably in the context of the loader: How can we make references a first class citizen instead of relying different parts of the code having to create them (and repeat the way they do?)

PR #44, opened 2025-02-27
feat: added base telemetry support in the rudder-cli

abhimanyubabbar:
## Description of the change
We need telemetry support in the cli to understand the usage by the customers. For initial version, we will be collecting:

1. Which command the user ran
2. What's the outcome of the command
3. Any specific flags which were added to the command run

Changes in the PR:
- [x] Capture the new variables needed to successfully execute telemetry through viper correctly.
- [x] Introduced a new telemetry command with subcommands `enable`, `disable` and `status`
- [x] Start firing `CLI Command Executed` event on every customer initiated command and capture the status of run in the event properties.

TBD:
Do we need any machine info to see usage across architectures ?

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


abhimanyubabbar:
One thing I wanna mention is that we also wanna capture osinfo in the track event payload coming to the upstream to us. I want to handle the OS setting requirement in a subsequent PR to make sure we handle it correctly.
cc @fxenik

cli/internal/cmd/telemetry/utils.go
```diff
@@ -0,0 +1,28 @@
+package telemetry
+
+import "github.com/rudderlabs/rudder-iac/cli/internal/telemetry"
+
+const (
+	CommandExecutedEvent = "Command Executed"
```
abhimanyubabbar, line 6: Hey @fxenik , do you think this event definition looks good ? Or should I change it to `CLI Command Executed` ?

fxenik: Command Executed sounds fine, but if you think we need to add CLI in front from extra clarity let's do it

cli/internal/config/config.go
```diff
@@ -13,7 +13,11 @@ import (
 	"github.com/tidwall/sjson"
 )
 
-var log = logger.New("config")
+var (
+	log                   = logger.New("config")
+	TelemetryWriteKey     = ""
```
fxenik, line 18: Let's add comment on top of this to let everyone know that is going to be populated through ldflags, in case someone moves that and forgets to change the paths.

cli/internal/config/config.go
```diff
@@ -53,9 +61,13 @@ func InitConfig(cfgFile string) {
 	viper.SetDefault("debug", false)
 	viper.SetDefault("verbose", false)
 	viper.SetDefault("apiURL", client.BASE_URL_V2)
+	viper.SetDefault("telemetry.writeKey", "")
```
fxenik, line 64: Why not set it to the `TelemetryWriteKey` var?
That way you can also avoid the check in the function that falls back to this var anyway,

cli/internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,102 @@
...
+	config.SetTelemetryDisabled(telemetryDisabled)
+}
+
+func EnableTelemetry() {
+	telemetryDisabled := false
+	config.SetTelemetryDisabled(telemetryDisabled)
+}
+
+func track(event string, properties analytics.Properties) error {
+	if telemetryDisabled {
```
fxenik, line 60: Why not just `if config.Telemetry.Disabled` ?

cli/internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,102 @@
...
+	client, err := analytics.NewWithConfig(writeKey, analytics.Config{
+		DataPlaneUrl: dataplaneURL,
+		Logger:       NewTelemetryLogger(),
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to create analytics client: %w", err)
+	}
+
+	defer client.Close()
```
fxenik, line 78: Shouldn't we be having a singleton client and use that instead of starting a new client on every track call?

cli/internal/cmd/auth.go
```diff
@@ -18,7 +19,14 @@ var loginCmd = &cobra.Command{
 	Use:   "login",
 	Short: "Login with an access token",
 	Args:  cobra.NoArgs,
-	Run: func(cmd *cobra.Command, args []string) {
-		auth.Login()
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+
+		defer func() {
```
fxenik, line 25: Maybe we run this in the root command's PersistentPostRun as a generic way to track all commands? Is it difficult to separate flags from flag arguments that way?

fxenik (changes requested):

cli/internal/telemetry/telemetry.go
abhimanyubabbar: I did think about this but in the end for now we are tracking a single command's result, so atmost we will be firing a single track call per execution. That was the understanding and also to keep things simple, went with a client created per call.

cli/internal/cmd/auth.go
abhimanyubabbar: I earlier added the command there but cobra doesn't guarantee to run the PostRun command on failure of Run command. Therefore, I bundled the track call within the defer func.

cli/internal/telemetry/telemetry.go
abhimanyubabbar: Telemetry can also be disabled through an env flag with gets read on the Initialise command in L36. So config is not the only way to disable the telemetry and this might be used in the github action where we don't wanna run the disable command everytime on the cli and can control it through the ENV.

Let me know if my approach makes sense to you ?

cli/internal/config/config.go
abhimanyubabbar: Will do !

cli/internal/cmd/telemetry/utils.go
abhimanyubabbar: I think CLI Command Executed might have better context in the downstream destination.

cli/internal/telemetry/telemetry.go
fxenik: The idea that config is used as the source of truth of configuration for the components, and it's using viper to merge local file configurations with configurations coming from envs. This is how the access token is handled as well. It can be set either way, but [read directly from config ](https://github.com/rudderlabs/rudder-iac/blob/18fda4c48382b4a0da49639780f95499bc270e98/cli/internal/app/dependencies.go#L52).

The only problem with this is if we initialise the telemetry component before we initialize config, but in this case I understand we initialize it afterwards.

cli/internal/telemetry/telemetry.go
fxenik: Cool, we can revisit if we need to introduce multiple events per command at some point.

cli/internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,102 @@
...
+	once.Do(func() {
+		conf := config.GetConfig()
+
+		v = version
+		if conf.Telemetry.Disabled {
+			telemetryDisabled = true
+			return
+		}
+
+		disabled := getEnvWithFallback("RUDDER_CLI_TELEMETRY_DISABLED", "")
```
fxenik, line 36: New comment: I think we should prefix the env with `RUDDERSTACK` instead of `RUDDER_CLI` so that all envs have a common prefix. `RUDDERSTACK_CLI_TELEMETRY_DISABLED` works as well.

fxenik (changes requested):

cli/internal/telemetry/telemetry.go
abhimanyubabbar: Make sense, I'll refactor this then.

cli/internal/config/config.go
abhimanyubabbar: Thanks for this, oversight on my part.

fxenik (approved):

PR #56, opened 2025-03-14
fix: docker images are built on tag pushes, instead of releases

fxenik:
## Description of the change

This is because GitHub doesn't trigger the docker workflow from the actions of another (goreleaser) flow.

From GitHub documentation:
> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur. For more information, see Automatic token authentication.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,9 +1,10 @@
 name: Docker Build and Push
 
 on:
-  release:
-    types: [published]
   push:
+    # Run on version tags and main branch
+    tags:
+      - "v*"
     branches:
       - main
   pull_request:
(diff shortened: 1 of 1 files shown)


abhimanyubabbar (approved):

abhimanyubabbar:
I approved the PR but if I understand correctly we are building the docker images on the tag push rather than a new release being published. This is due to the inherent nature of the GITHUB_TOKEN which is unable to spin off new workflows.

My question is should we think about using a different token which allows us to spin this worklow on type: released or is it safe to be kicked off on every tag push ? I mean wouldn't it be easy push to the same tag by deleting and retagging ?

PR #64, opened 2025-05-13
chore: added validation rules for the custom types

abhimanyubabbar:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/cli/pkg/localcatalog/tracking_plan.go
+++ b/cli/pkg/localcatalog/tracking_plan.go
@@ -7,9 +7,10 @@ import (
 )
 
 var (
-	PropRegex    = regexp.MustCompile(`^#\/properties\/(.*)\/(.*)$`)
-	EventRegex   = regexp.MustCompile(`^#\/events\/(.*)\/(.*)$`)
-	IncludeRegex = regexp.MustCompile(`^#\/tp\/(.*)\/event_rule\/(.*)$`)
+	PropRegex       = regexp.MustCompile(`^#\/properties\/(.*)\/(.*)$`)
+	EventRegex      = regexp.MustCompile(`^#\/events\/(.*)\/(.*)$`)
+	IncludeRegex    = regexp.MustCompile(`^#\/tp\/(.*)\/event_rule\/(.*)$`)
+	CustomTypeRegex = regexp.MustCompile(`^#\/custom-types\/(.*)\/(.*)$`)
 )
 
 type CatalogResourceFetcher interface {
--- a/cli/pkg/validate/refs.go
+++ b/cli/pkg/validate/refs.go
@@ -2,6 +2,7 @@ package validate
 
 import (
 	"fmt"
+	"strings"
 
 	catalog "github.com/rudderlabs/rudder-iac/cli/pkg/localcatalog"
 )
@@ -55,6 +56,34 @@ func (rv *RefValidator) Validate(dc *catalog.DataCatalog) []ValidationError {
 							error:     fmt.Errorf("property reference '%s' at index %d not found in catalog", prop.ID, i),
 						})
 					}
+				}
+			}
+		}
+	}
+
+	// Validate property type custom type references
+	for group, props := range dc.Properties {
+		for _, prop := range props {
+			reference := fmt.Sprintf("#/properties/%s/%s", group, prop.LocalID)
+
+			// Check if the type field contains a custom type reference
+			if strings.HasPrefix(prop.Type, "#/custom-types/") {
+				matches := catalog.CustomTypeRegex.FindStringSubmatch(prop.Type)
+				if len(matches) != 3 {
+					errs = append(errs, ValidationError{
+						Reference: reference,
+						error:     fmt.Errorf("custom type reference in type field has invalid format. Should be '#/custom-types/<group>/<id>'"),
+					})
+					continue
+				}
+
+				// Validate custom type existence
+				customTypeGroup, customTypeID := matches[1], matches[2]
+				if customType := dc.CustomType(customTypeGroup, customTypeID); customType == nil {
+					errs = append(errs, ValidationError{
+						Reference: reference,
+						error:     fmt.Errorf("custom type reference '%s' not found in catalog", prop.Type),
+					})
 				}
 			}
 		}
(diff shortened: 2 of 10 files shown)


cli/pkg/localcatalog/tracking_plan.go
```diff
@@ -7,9 +7,10 @@ import (
 )
 
 var (
-	PropRegex    = regexp.MustCompile(`^#\/properties\/(.*)\/(.*)$`)
-	EventRegex   = regexp.MustCompile(`^#\/events\/(.*)\/(.*)$`)
-	IncludeRegex = regexp.MustCompile(`^#\/tp\/(.*)\/event_rule\/(.*)$`)
+	PropRegex       = regexp.MustCompile(`^#\/properties\/(.*)\/(.*)$`)
+	EventRegex      = regexp.MustCompile(`^#\/events\/(.*)\/(.*)$`)
+	IncludeRegex    = regexp.MustCompile(`^#\/tp\/(.*)\/event_rule\/(.*)$`)
+	CustomTypeRegex = regexp.MustCompile(`^#\/custom-types\/(.*)\/(.*)$`)
```
Jayachand, line 13: name should start with capital letter

fxenik: You mean the custom type's name as created in the backend? Is this a constraint we enforce already in our APIs?
For reference this regex is for references in YAML files, which require the custom type's local id, not the name. So this type of validation should be implemented in a different place, where we validate the name property in the custom type spec.

cli/pkg/validate/refs.go
```diff
@@ -25,6 +28,67 @@ func (rv *RefValidator) Validate(dc *catalog.DataCatalog) []ValidationError {
...
+				customTypeGroup, customTypeID := matches[1], matches[2]
+				if customType := dc.CustomType(customTypeGroup, customTypeID); customType == nil {
+					errs = append(errs, ValidationError{
+						Reference: reference,
+						error:     fmt.Errorf("custom type reference '%s' not found in catalog", prop.Type),
+					})
+				}
+			}
+		}
+	}
```
fxenik, line 90: This is mostly a mental note, doesn't necessarily require any changes yet:
Looking for valid references (valid reference string, existence of referred resources) looks like a common pattern, which ideally could be abstracted to:
- handle common parsing logic, and resource lookup.
- have common consistent error reporting.

Lookup should be possible even at the YAML or generic resource representation, not necessarily depend on the parsed local catalog state the provider maintains. That would require significant changes though, that's why I'm not expecting any changes atm. It's important to keep in mind though because we're probably going to have to support cross-domain references e.g retl sql models to an account, or source to a tracking plan.

cli/pkg/validate/required_keys.go
```diff
@@ -92,5 +112,247 @@ func (rk *RequiredKeysValidator) Validate(dc *catalog.DataCatalog) []ValidationE
...
+			if customType.Type == "object" {
+				for i, prop := range customType.Properties {
+					if prop.ID == "" {
+						errors = append(errors, ValidationError{
+							error:     fmt.Errorf("id field is mandatory for property at index %d in custom type", i),
+							Reference: reference,
+						})
+					}
+				}
+			}
```
fxenik, line 155: This is going to be replaced with a check for "$ref", right?

cli/pkg/validate/required_keys.go
```diff
@@ -92,5 +112,247 @@ func (rk *RequiredKeysValidator) Validate(dc *catalog.DataCatalog) []ValidationE
...
+				errors = append(errors, ValidationError{
+					error:     fmt.Errorf("custom type name must start with a capital letter and contain only letters, numbers, underscores and dashes, and be between 2 and 65 characters long"),
+					Reference: reference,
+				})
+			}
+
+			// Type validation
+			if !slices.Contains(validTypes, customType.Type) {
+				errors = append(errors, ValidationError{
+					error:     fmt.Errorf("invalid data type. Acceptable values are string, number, integer, boolean, null, array, and object"),
```
fxenik, line 168: Does it make sense to construct the error message using the valid format arrays, so as it to make sure we don't have to keep those two in sync going forward?

cli/pkg/validate/required_keys.go
```diff
@@ -92,5 +112,247 @@ func (rk *RequiredKeysValidator) Validate(dc *catalog.DataCatalog) []ValidationE
...
+					Reference: reference,
+				})
+			}
+
+			// Type-specific config validations
+			if customType.Config != nil {
```
fxenik, line 174: Should we be able to apply the same validations in properties' propConfig?

fxenik (changes requested):

cli/pkg/validate/required_keys.go
abhimanyubabbar: Yes

cli/pkg/validate/required_keys.go
abhimanyubabbar: Yes, you are right. Dumb LLM mistake 😮‍💨

cli/pkg/validate/required_keys.go
abhimanyubabbar: It's a good point, I'll verify and fix it.

cli/pkg/validate/refs.go
abhimanyubabbar: Yes, you are right. This is why the current validation logic is flawed and limited. There are common validations check which doesn't need complete parsing of the files.

cli/pkg/validate/required_keys.go
abhimanyubabbar: @fxenik although I am tempted to use the same config check for properties as well, does it make sense to do that in a separate PR to separate the concerns ?

If yes, I can create another PR to handle this change in a more DRY friendly way.

fxenik: Yes we don't have to do it now if you think it needs more thought. I was just thinking about what we discussed in the past, that the CLI should be able to validate propConfig even before making any API calls. We can have this as a backlog item and see if we can reuse the type checking that's introduced here.

cli/pkg/validate/required_keys.go
fxenik: haha that's why we need human reviewers in the loop :p

fxenik (approved):

PR #72, opened 2025-05-19
chore: add references in property type to custom types

abhimanyubabbar:
## Description of the change

Type of property can now additionally hold references of custom types which can be changed / updated by the user.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


cli/pkg/provider/state/customtype.go
```diff
@@ -178,7 +178,7 @@ func (s *CustomTypeState) FromResourceData(from resources.ResourceData) {
 	s.Description = MustString(from, "description")
 	s.Type = MustString(from, "type")
 	s.Config = MapStringInterface(from, "config", make(map[string]any))
-	s.Version = MustInt(from, "version")
+	s.Version = int(MustFloat64(from, "version"))
```
fxenik, line 181: Why is this casting necessary? Why would version be a float64?

cli/internal/cmd/trackingplan/apply/apply.go
```diff
@@ -155,6 +155,20 @@ func createResourceGraph(catalog *localcatalog.DataCatalog) *resources.Graph {
 				Config:      prop.Config,
 			}
 
+			if strings.HasPrefix(prop.Type, "#/custom-types/") {
+				customTypeURN := getURNFromRef(prop.Type)
+
+				if customTypeURN != "" {
+					log.Debug("property references custom type", "property", prop.LocalID, "customTypeRef", prop.Type, "customTypeURN", customTypeURN)
+					args.Type = resources.PropertyRef{
```
fxenik, line 163: Having Type as any is troubling, especially if the actually type doesn't depend on the actual property configurations, it depends at which point we parse it. We start by expecting either a string type (with a simple property type) or a custom-type reference, but later  (during create and update) we treat the same args as having either the simple property type or the custom-type's name.

Post this version, (or as part of making custom types support GA) we need to revisit how we handle args, state and resource data.

cli/pkg/provider/state/customtype.go
abhimanyubabbar: That's how go works, anything number added to interface{} will be typecasted by default into Float64.

cli/internal/cmd/trackingplan/apply/apply.go
abhimanyubabbar: Yeah I agree. But wouldn't this be true of any reference. When we create the entity to hold a reference it needs to be of any type right ?
To respect both the resolved type and PropertyRef type. I wasn't able to find a better way to do conversion that this and would like your opinion on how can we improve it.

Jayachand (approved):

PR #74, opened 2025-05-21
fix: fixed the process of upserting trackingplan with the property which have custom types correctly referenced

abhimanyubabbar:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/cli/pkg/provider/state/trackingplan.go
+++ b/cli/pkg/provider/state/trackingplan.go
@@ -1,6 +1,9 @@
 package state
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/rudderlabs/rudder-iac/cli/internal/syncer/resources"
 	"github.com/rudderlabs/rudder-iac/cli/pkg/localcatalog"
 )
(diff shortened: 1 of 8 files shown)


cli/pkg/provider/state/trackingplan.go
```diff
@@ -243,24 +248,75 @@ func (args *TrackingPlanPropertyArgs) Diff(other *TrackingPlanPropertyArgs) bool
...
-func (args *TrackingPlanArgs) FromCatalogTrackingPlan(from *localcatalog.TrackingPlan) {
+func (args *TrackingPlanPropertyArgs) FromCatalogTrackingPlanEventProperty(prop *localcatalog.TPEventProperty, urnFromRef func(string) string) error {
+	args.Name = prop.Name
+	args.Description = prop.Description
+	args.LocalID = prop.LocalID
+	args.Required = prop.Required
+	args.Type = prop.Type
+	args.Config = prop.Config
+	args.HasCustomTypeRef = false
+	args.HasItemTypesRef = false
```
fxenik, line 259: @abhimanyubabbar the problem we discussed earlier today now is starting to become clearer to me.
We don't necessarily have to fix now, because the solution might be to complex). One way to think about it is to change the Syncer's dereference implementation to not replace the PropertyRefs with a resolved value, replace it instead with a ResolvedPropertyRef which maintains both the Reference properties (so that the provider can handle that accordingly) **and** the resolved value. Something along these lines anyway, we can figure out the details afterwards.

For now, we need to consider that diffs in the resource data are going to be reported to the user as information of what has been updated. I'm not sure how these new Has* properties are going to manifest in this case. It might be a bit confusing, but then again the whole diff view is currently confusing as is. I feel it's fine for now if product doesn't think it's a big deal for the scope of this project.

Jayachand (dismissed):

Jayachand (approved):

cli/pkg/provider/state/trackingplan.go
abhimanyubabbar: You are right to point that diffs will show up these values and this is what I was thinking to avoid earlier as well. The HasProperties are tightly linked with the customtypes changes, given the values don't change, they also don't change

fxenik (approved):

PR #80, opened 2025-06-10
feat: add property name validation for leading/trailing whitespace (RUD-2347)

dhawal1248:
Implements validation in CLI to prevent property names with leading and trailing whitespace characters, aligning with backend behavior. Adds comprehensive test coverage and maintains backward compatibility. Resolves RUD-2347.


cli/pkg/validate/required_keys.go
```diff
@@ -46,6 +46,14 @@ func (rk *RequiredKeysValidator) Validate(dc *catalog.DataCatalog) []ValidationE
 				})
 			}
 
+			// Validate property name doesn't have leading or trailing whitespace
+			if prop.Name != "" && prop.Name != strings.TrimSpace(prop.Name) {
```
abhimanyubabbar, line 50:
```suggestion
	if prop.Name != strings.TrimSpace(prop.Name) {
```
Should this check be enough ?

cli/pkg/validate/required_keys_test.go
```diff
@@ -356,3 +357,189 @@ func Test_IsNumber(t *testing.T) {
...
+						Description: "Property with trailing space",
+						Type:        "string",
+					},
+					{
+						LocalID:     "valid-prop",
+						Name:        "Valid Property",
+						Description: "Valid property",
+						Type:        "string",
+					},
+				},
```
abhimanyubabbar, line 510: I would rather have this single test case and remove all the others. They add more noise, whatever we are testing is contained within this set of inputs.

PR #85, opened 2025-06-12
chore: setup the base for e2e tests with binary building and config setup

abhimanyubabbar:
## Description of the change

In this PR, we implement the base setup for the e2e tests with:
1. Ability to execute commands using the Executor interface
2. Setting a cli binary from scratch and placing it in PATH
3. Ability to create config on demand for test
4. A simple E2E setup bringing all these things together

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


cli/tests/config.go
```diff
@@ -0,0 +1,125 @@
...
+//	        Build()
+//
+// The Build method returns the path to the generated config file, a cleanup
+// function, and an error if something went wrong.
+//
+// All methods mutate the builder in-place and return the same instance,
+// allowing chaining.
+// NOTE: Only the fields relevant to current tests are implemented. Extend as
+// needed to cover additional configuration keys.
+type ConfigBuilder struct {
```
dhawal1248, line 30: can we create ConfigBuilder as a composite type that consits of the actual Config struct and the filename
```suggestion
type ConfigBuilder struct {
    Config
    filename string
}
````

Comment (dhawal1248) [reply cli/tests/config.go:30]: We could add builder functions like `WithConfig` as and when required by the tests

File: cli/tests/helpers.go [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+// The returned []byte contains stdout and stderr combined, identical to exec.Cmd.CombinedOutput().
+// The error will be non-nil if the command exits with non-zero status or fails to start.
+// Callers may use errors.As(err, *exec.ExitError) to inspect exit code if required.
+//
+// Example:
+//
+//	exec := tests.NewCmdExecutor("")
+//	out, err := exec.Execute("echo", "hello")
+//	fmt.Printf("%s", out) // prints "hello\n"
+type Executor interface {
```
dhawal1248, line 24: can we use the exec package directly here without adding the Executor abstraction? Will we support different types of executors? I feel we would mostly need just a os command executor..we can expose a helper function called `ExecuteCommand` or something..this feels simpler to me

cli/tests/config.go
fxenik: I would actually try to keep things simple here, only include config properties that we actually want to include during e2e tests. Otherwise, the ConfigBuilder will start to replicate work that would feel more natural to the config package itself, i.e creating a config file (or its contents) from Config data.

I would also like to mention here again that I would prefer the access token (and maybe api url as well) out of the config file if we're already expecting those to be passed through environmental variables to the test suite. We don't need to test that config is parsed correctly by the CLI, that can be done with unit tests.

cli/tests/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+	path, err := bin.Setup()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	oldpath := os.Getenv("PATH")
+	binDir := filepath.Dir(path)
+
+	os.Setenv("PATH", fmt.Sprintf("%s:%s", binDir, oldpath))
```
fxenik, line 44: Do we need to put the binary in the PATH? Can we just rely on the absolute path of the binary?

cli/tests/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	defer os.Setenv("PATH", oldpath)
+
+	exitCode := m.Run()
+
+	_ = bin.Clean()
+	os.Exit(exitCode)
```
fxenik, line 51: Should Clean run in a defer function?

cli/tests/helpers.go
abhimanyubabbar: Testing becomes much easier when you have an interface for Executor. You can use a test implementation and inject this mocked dependency when need arises in future.

cli/tests/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	defer os.RemoveAll(tmpDir)
+
+	exec, err := NewCmdExecutor("")
+	if err != nil {
+		fmt.Println("failed to init executor:", err)
+		os.Exit(1)
+	}
+
+	bin, err := NewCLIBinary(tmpDir, exec)
```
fxenik, line 29: Do we need tmpDir passed as argument given that NewCLIBinary can already handle that? Isn't it better for the CLIBinary being responsible for both creating temp directories/files as well as cleaning them up?

fxenik (changes requested):

cli/tests/main_test.go
abhimanyubabbar: Earlier temp directory creation was supposed to shared between binary and config files. I envisioned a flow where once I have a binary, I can have multiple locations of the config file and then provide the binary which config file location to use when running a e2e test. 

Let me think and refactor this flow a bit

cli/tests/config.go
```diff
@@ -0,0 +1,125 @@
...
+	telemetryWriteKey  string
+	telemetryDataplane string
+	telemetryAnonymous string
+
+	filePath string
+}
+
+func NewConfigBuilder(path string) *ConfigBuilder {
+	apiURL := os.Getenv("CONFIG_BACKEND_URL")
+	accessToken := os.Getenv("ACCESS_TOKEN")
```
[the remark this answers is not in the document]
abhimanyubabbar: Let me make change to it

PR #93, opened 2025-06-16
chore: use PR number to build docker images for PRs

dhawal1248:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -37,14 +37,14 @@ jobs:
           else
             # For PR, use PR branch name
             if [[ ${{ github.event_name }} == 'pull_request' ]]; then
-              PR_TAG="pr-${{ github.event.pull_request.number }}"
+              TAG="pr-${{ github.event.pull_request.number }}"
             else
               # For main branch push
-              PR_TAG=$GITHUB_REF_NAME
+              TAG=$GITHUB_REF_NAME
             fi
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -37,14 +37,14 @@ jobs:
           else
             # For PR, use PR branch name
             if [[ ${{ github.event_name }} == 'pull_request' ]]; then
-              BRANCH=$GITHUB_HEAD_REF
+              PR_TAG="pr-${{ github.event.pull_request.number }}"
```
fxenik, line 40: calling this `PR_TAG` instead of `BRANCH` might be misleading because we don't populate just for PRs. Can we keep the original `BRANCH` variable name or just `TAG`?

fxenik (changes requested):

.github/workflows/docker.yml
dhawal1248: we would only reach this else block if it is a PR @fxenik

fxenik: Yeap but we're setting PR_TAG in the block after that as well, where the main branch is handled, right?

dhawal1248: ah got it, makes sense

fxenik (approved):

PR #95, opened 2025-06-17
feat(cli): show panics in console when debug flag is enabled

dhawal1248:
When the debug flag is enabled, panics are now shown in the console output in addition to being logged to cli.log. This improves developer experience and debugging.\n\n- Panics are always logged to cli.log\n- If debug is enabled, full panic details and stack trace are printed to the console\n- Non-debug mode retains the user-friendly error message\n\nImplements RUD-2395.

### DIFF
--- a/cli/internal/cmd/root.go
+++ b/cli/internal/cmd/root.go
@@ -25,10 +25,21 @@ var (
 
 func recovery() {
 	if r := recover(); r != nil {
-		fmt.Println(emoji.Sprintf("\n:skull:Oops! Unexpected error occurred. Please contact tech support.\n"))
+		// Always log to file
 		log.Error("panic detected", "error", r)
 		log.Error(string(debug.Stack()))
 
+		// If debug mode is enabled, show detailed panic info in console and exit
+		if viper.GetBool("debug") {
+			fmt.Println("\n🔍 Debug Mode: Panic Details")
+			fmt.Printf("Error: %v\n", r)
+			fmt.Println("\nStack Trace:")
+			fmt.Println(string(debug.Stack()))
+			os.Exit(1)
+		}
+
+		// In non-debug mode, show the simple error message and exit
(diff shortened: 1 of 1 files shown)


cli/internal/cmd/root.go
```diff
@@ -25,10 +25,21 @@ var (
 
 func recovery() {
 	if r := recover(); r != nil {
-		fmt.Println(emoji.Sprintf("\n:skull:Oops! Unexpected error occurred. Please contact tech support.\n"))
+		// Always log to file
 		log.Error("panic detected", "error", r)
 		log.Error(string(debug.Stack()))
 
+		// If debug mode is enabled, show detailed panic info in console and exit
+		if viper.GetBool("debug") {
```
fxenik, line 33: nit: this hasn't been followed consistently in all places where we access config parameters, but we want to move away from accessing them directly from viper, and use the config struct instead. Ideally we should do:
```
config.GetConfig().Debug
```

Maybe this is a good opportunity to fix it in other places as well where we use viper outside the config package

fxenik (changes requested):

cli/internal/cmd/root.go
```diff
@@ -25,10 +25,21 @@ var (
 
 func recovery() {
 	if r := recover(); r != nil {
-		fmt.Println(emoji.Sprintf("\n:skull:Oops! Unexpected error occurred. Please contact tech support.\n"))
+		// Always log to file
 		log.Error("panic detected", "error", r)
 		log.Error(string(debug.Stack()))
```
abhimanyubabbar, line 30: This panic log is not properly formatted which means it won't be formatted properly on the UI to the customers as well. Should we fix the formatting on the logged stack trace ?

abhimanyubabbar (changes requested):

cli/internal/cmd/root.go
dhawal1248: sure, what kind of formatting were you thinking of?****

abhimanyubabbar (approved):

fxenik (approved):

PR #96, opened 2025-06-19
chore: auto-append /v2 to API URLs (RUD-2384)

dhawal1248:
Fixes RUD-2384 - Users can now set RUDDERSTACK_API_URL without /v2 suffix. API client automatically appends /v2 to all endpoints. Resolves: RUD-2384

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -24,7 +24,7 @@ type Client struct {
 	Connections  *connections
 }
 
-const BASE_URL_V2 = "https://api.rudderstack.com"
+const BASE_URL = "https://api.rudderstack.com"
 const API_VERSION = "v2"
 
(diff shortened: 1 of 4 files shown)


api/client/client.go
```diff
@@ -24,7 +24,8 @@ type Client struct {
 	Connections  *connections
 }
 
-const BASE_URL_V2 = "https://api.rudderstack.com/v2"
+const BASE_URL_V2 = "https://api.rudderstack.com"
```
abhimanyubabbar, line 27: We don't call it `BASE_URL_V2` then right ?
Just `BASE_URL` would suffice

abhimanyubabbar (changes requested):

PR #101, opened 2025-07-01
feat: api client for retl sources

am6010:
## Description of the change

This PR implements a new API client for RETL (Real-time ETL) sources, providing a complete interface for managing RETL sources and their states. The implementation follows the same patterns as the existing catalog client.

Key additions:
- CRUD operations for RETL sources (create, read, update, delete)
- State management operations (read state, update resource state)
- Comprehensive error handling and testing
- Integration with existing client infrastructure

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/api/client/retl/types.go
+++ b/api/client/retl/types.go
@@ -27,34 +27,34 @@ type PutStateRequest struct {
 
 // RETLSource represents a RETL source in the API
 type RETLSource struct {
-	ID                   string           `json:"id,omitempty"`
-	Name                 string           `json:"name"`
-	Config               RETLSourceConfig `json:"config"`
-	IsEnabled            bool             `json:"enabled,omitempty"`
-	SourceType           string           `json:"sourceType"`
-	SourceDefinitionName string           `json:"sourceDefinitionName"`
-	AccountID            string           `json:"accountId"`
-	CreatedAt            *time.Time       `json:"createdAt,omitempty"`
-	UpdatedAt            *time.Time       `json:"updatedAt,omitempty"`
+	ID                   string             `json:"id"`
+	Name                 string             `json:"name"`
+	Config               RETLSQLModelConfig `json:"config"`
+	IsEnabled            bool               `json:"enabled"`
+	SourceType           string             `json:"sourceType"`
+	SourceDefinitionName string             `json:"sourceDefinitionName"`
+	AccountID            string             `json:"accountId"`
+	CreatedAt            *time.Time         `json:"createdAt"`
+	UpdatedAt            *time.Time         `json:"updatedAt"`
 }
 
 type RETLSourceCreateRequest struct {
-	Name                 string           `json:"name"`
-	Config               RETLSourceConfig `json:"config"`
-	SourceType           string           `json:"sourceType"`
-	SourceDefinitionName string           `json:"sourceDefinitionName"`
-	AccountID            string           `json:"accountId"`
+	Name                 string             `json:"name"`
+	Config               RETLSQLModelConfig `json:"config"`
+	SourceType           string             `json:"sourceType"`
+	SourceDefinitionName string             `json:"sourceDefinitionName"`
+	AccountID            string             `json:"accountId"`
 }
 
 type RETLSourceUpdateRequest struct {
-	Name      string           `json:"name"`
-	Config    RETLSourceConfig `json:"config"`
-	IsEnabled bool             `json:"enabled,omitempty"`
-	AccountID string           `json:"accountId"`
+	Name      string             `json:"name"`
+	Config    RETLSQLModelConfig `json:"config"`
+	IsEnabled bool               `json:"enabled"`
+	AccountID string             `json:"accountId"`
 }
 
 // RETLSourceConfig represents the config of a RETL SQL model source
-type RETLSourceConfig struct {
+type RETLSQLModelConfig struct {
 	PrimaryKey  string `json:"primaryKey"`
 	Sql         string `json:"sql"`
 	Description string `json:"description,omitempty"`
--- a/api/client/retl/sources_test.go
+++ b/api/client/retl/sources_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestCreateRetlSource(t *testing.T) {
-	sourceConfig := retl.RETLSourceConfig{
+	sourceConfig := retl.RETLSQLModelConfig{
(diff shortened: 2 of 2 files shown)


api/client/retl/retl.go
```diff
@@ -0,0 +1,75 @@
...
+
+// IsRETLNotFoundError checks if an error is a "not found" error
+func IsRETLNotFoundError(err error) bool {
+	var apiErr *client.APIError
+
+	if ok := errors.As(err, &apiErr); !ok {
+		return false
+	}
+	return apiErr.HTTPStatusCode == 404 || (apiErr.HTTPStatusCode == 400 && strings.Contains(apiErr.Message, "not found"))
+}
```
fxenik, line 65: Is this necessary for RETL? That was a workaround for data catalog APIs because they were responding with 400 in some cases were resources are not found. With RETL APIs, being new, we can make sure 404 is always returned when a resource is not found.

In general, we want to avoid the pattern of checking the error message contents to identify different classes of errors. It's better if we rely on status codes first, and then maybe on specific error codes in the responses if additional information is needed (but let's make sure we have a need for this first).

api/client/retl/retl.go
```diff
@@ -0,0 +1,75 @@
...
+
+// IsRETLAlreadyExistsError checks if an error is an "already exists" error
+func IsRETLAlreadyExistsError(err error) bool {
+	var apiErr *client.APIError
+
+	if ok := errors.As(err, &apiErr); !ok {
+		return false
+	}
+	return apiErr.HTTPStatusCode == 400 && strings.Contains(apiErr.Message, "already exists")
+}
```
fxenik, line 75: In which cases do we have this kind of conflict, e.g do we consider the name unique? In those cases, I think it's better to return a 409 Conflict status code for conflicts with existing workspace data.

Let's double check how our APIs behave in case of conflicts, and add a backlog task to change that response code if necessary.

api/client/retl/sources.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// CreateRetlSource creates a new RETL source
+func (r *RudderRETLStore) CreateRetlSource(ctx context.Context, source *RETLSourceCreateRequest) (*RETLSource, error) {
+	data, err := json.Marshal(source)
+	if err != nil {
+		return nil, fmt.Errorf("marshalling source: %w", err)
+	}
+
+	resp, err := r.client.Do(ctx, "POST", "/retl-sources", bytes.NewReader(data))
```
fxenik, line 17: I made a change in my recent PR for the Account API client, where I'm not adding the /v2 prefix by default on all paths. The reason is that there was a bug when pagination was used (it is introduced in accounts APIs) where the /v2 was also added to the `next` url that server was responding, which already included v2. I also think it's better if we explicitly add the version on the endpoints in case we want to introduce other versions in the future.

That PR is now merged, so let's rebase this one on main, and add "/v2/" in front of those paths.

api/client/retl/types.go
```diff
@@ -0,0 +1,70 @@
...
+	IsEnabled            bool             `json:"enabled,omitempty"`
+	SourceType           string           `json:"sourceType"`
+	SourceDefinitionName string           `json:"sourceDefinitionName"`
+	AccountID            string           `json:"accountId"`
+	CreatedAt            *time.Time       `json:"createdAt,omitempty"`
+	UpdatedAt            *time.Time       `json:"updatedAt,omitempty"`
+}
+
+type RETLSourceCreateRequest struct {
+	ID                   string           `json:"id,omitempty"`
```
fxenik, line 43: Do we need ID in the create request?

api/client/retl/types.go
```diff
@@ -0,0 +1,70 @@
...
+	Name                 string           `json:"name"`
+	Config               RETLSourceConfig `json:"config"`
+	SourceType           string           `json:"sourceType"`
+	SourceDefinitionName string           `json:"sourceDefinitionName"`
+	AccountID            string           `json:"accountId"`
+}
+
+type RETLSourceUpdateRequest struct {
+	ID        string           `json:"id,omitempty"`
```
fxenik, line 52: Can this actually be empty in the update request?

api/client/retl/types.go
```diff
@@ -0,0 +1,70 @@
...
+	Name      string           `json:"name"`
+	Config    RETLSourceConfig `json:"config"`
+	IsEnabled bool             `json:"enabled,omitempty"`
+	AccountID string           `json:"accountId"`
+	SourceID  string           `json:"sourceId"`
```
fxenik, line 57: Why do we need a SourceID here?

api/client/retl/types.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// RETLSourceConfig represents the config of a RETL SQL model source
+type RETLSourceConfig struct {
```
fxenik, line 61: Let's call this `RETLSQLModelConfig` for now. We don't have to support other source types for now, and when we do can probably introduce other create/update functions that expect different inputs for tables, profiles etc. It's better though if we don't name this as a generic source config type, to not misinterpret this as the common struct for all possible config schemas.

api/client/retl/sources.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// CreateRetlSource creates a new RETL source
+func (r *RudderRETLStore) CreateRetlSource(ctx context.Context, source *RETLSourceCreateRequest) (*RETLSource, error) {
+	data, err := json.Marshal(source)
```
fxenik, line 12: I see in some cases (e.g create) we marshal the request directly, while in others (e.g update, put state) we populate a map from the request and marshal that instead. It's better if we have one approach for every case where we need to create a request body. Either serialize the request (relying on json tags) or directly have the client functions populate the bodies from the request arguments.

fxenik (changes requested):

api/client/retl/retl.go
am6010: My bad we don't need this helper function

api/client/retl/retl.go
am6010: I checked the retl module and didn't find any obvious conflict cases. I have removed this function.

fxenik (approved):

PR #102, opened 2025-07-02
chore: merge pkg/provider into internal/providers

dhawal1248:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


cli/internal/providers/datacatalog/provider.go
```diff
@@ -25,7 +24,7 @@ type Provider struct {
 
 func New(client catalog.DataCatalog) *Provider {
 	return &Provider{
-		SyncProvider: provider.NewCatalogProvider(client),
+		SyncProvider: NewCatalogProvider(client),
```
fxenik, line 27: Do you think it's still necessary to have to different structs for the data catalog provider, this Provider struct and the one returned by NewCatalogProvider? It was implemented like this before to avoid conflicts during development (hence the composition) but maybe there's no value in keeping both now, with their own New functions, their own client variables etc.

If we're keeping them we might consider renaming the older one to something more sync specific (it might be confusing to name one as CatalogProvider and the other just Provider), and consider making the NewCatalogProvider private to the package.

dhawal1248: makes sense, @fxenik made the required changes

fxenik (approved):

PR #161, opened 2025-09-03
chore: improve error handling for 5xx API errors

shekhar-rudder:
**Description**
When the API returned a 504 error, the CLI showed a confusing message: `could not parse error response from API: invalid character '<' looking for beginning of value`. This PR improves that by surfacing a clearer, human-friendly error instead.

**Changes**
- Detect 5xx responses and return a generic message: “Something went wrong. Try again shortly.” instead of attempting to parse the server response

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -87,6 +87,15 @@ func (c *Client) Do(ctx context.Context, method, path string, body io.Reader) ([
 		return nil, err
 	}
 
+	// For 5xx errors, return a generic message instead of parsing the response
+	if res.StatusCode >= 500 && res.StatusCode <= 599 {
+		return nil, &APIError{
+			HTTPStatusCode: res.StatusCode,
+			Message: "Something went wrong. Try again shortly.",
+		}
+	}
+	
+
 	// check if response has an error status code and parse API error accordingly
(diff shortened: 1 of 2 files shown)


api/client/client.go
```diff
@@ -87,6 +87,15 @@ func (c *Client) Do(ctx context.Context, method, path string, body io.Reader) ([
 		return nil, err
 	}
 
+	// For 5xx errors, return a generic message instead of parsing the response
+	if res.StatusCode >= 500 && res.StatusCode <= 599 {
+		return nil, &APIError{
+			HTTPStatusCode: res.StatusCode,
+			Message: "Something went wrong. Try again shortly.",
+		}
+	}
```
abhimanyubabbar, line 96: We can do this but my worry is we will gobble up the actual data payload that contained the reason for the issue. 

The error could be logged in the `cli.log` file for further debugging in future and hence should contain all the necessary upstream error information.

Wdyt ?

abhimanyubabbar: What we can do is try and represent 2 kinds of error messages in the APIError maybe, one a user consumable and another one which is actual error message. This can then be caller's responsibility to check and display error appropriately.

PR #169, opened 2025-09-15
feat: orchestrator for typer code generation

akashrpo:
- Introduces a new orchestrator that coordinates RudderTyper generation workflow:
  - Fetch tracking plan data (mock-backed in this PR).
  - Invoke platform generator (Kotlin supported).
  - Write generated files via FileManager with safe, atomic operations.

### DIFF
--- a/cli/internal/typer/generator/core/file_manager.go
+++ b/cli/internal/typer/generator/core/file_manager.go
@@ -22,7 +22,7 @@ func NewFileManager(baseDir string) *FileManager {
 
 // WriteFile writes a single file to the filesystem atomically
 // It creates parent directories as needed and handles file writing safely
-func (fm *FileManager) WriteFile(file File) error {
+func (fm *FileManager) WriteFile(file *File) error {
 	if err := fm.validateFile(file); err != nil {
 		return err
 	}
@@ -54,14 +54,14 @@ func (fm *FileManager) WriteFiles(files []*File) error {
 
 	// Validate all files first
 	for i, file := range files {
-		if err := fm.validateFile(*file); err != nil {
+		if err := fm.validateFile(file); err != nil {
 			return fmt.Errorf("file %d: %w", i, err)
 		}
 	}
 
 	// Write all files
 	for _, file := range files {
-		if err := fm.WriteFile(*file); err != nil {
+		if err := fm.WriteFile(file); err != nil {
 			return err
 		}
 	}
@@ -70,7 +70,7 @@ func (fm *FileManager) WriteFiles(files []*File) error {
 }
 
 // validateFile validates file input parameters
-func (fm *FileManager) validateFile(file File) error {
+func (fm *FileManager) validateFile(file *File) error {
 	if file.Path == "" {
 		return fmt.Errorf("file path cannot be empty")
 	}
--- a/cli/internal/typer/generator/core/file_manager_test.go
+++ b/cli/internal/typer/generator/core/file_manager_test.go
@@ -15,44 +15,44 @@ func TestFileManager_WriteFile(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		file           File
-		verifyDirPerms bool                                          // whether to verify directory permissions were set correctly
-		setupFunc      func(tempDir string, fm *FileManager) error   // setup before main test
-		verifyFunc     func(t *testing.T, tempDir string, file File) // custom verification logic
+		file           *File
+		verifyDirPerms bool                                           // whether to verify directory permissions were set correctly
+		setupFunc      func(tempDir string, fm *FileManager) error    // setup before main test
+		verifyFunc     func(t *testing.T, tempDir string, file *File) // custom verification logic
 	}{
 		"single file": {
-			file: File{
+			file: &File{
 				Path:    "test.txt",
 				Content: "Hello, World!",
 			},
 		},
 		"nested directories": {
-			file: File{
+			file: &File{
 				Path:    "nested/deep/dir/test.txt",
 				Content: "Nested content",
 			},
 			verifyDirPerms: true,
 		},
 		"file with special characters": {
-			file: File{
+			file: &File{
 				Path:    "special/file-name_with.chars.txt",
 				Content: "Content with special chars: 🎉 & symbols!",
 			},
 		},
 		"empty content": {
-			file: File{
+			file: &File{
 				Path:    "empty.txt",
 				Content: "",
 			},
 		},
 		"large file content": {
-			file: File{
+			file: &File{
 				Path:    "large.txt",
 				Content: strings.Repeat("A", 1024*1024), // 1MB
 			},
 		},
 		"multiline content": {
-			file: File{
+			file: &File{
 				Path: "multiline.txt",
 				Content: `Line 1
 Line 2
@@ -61,16 +61,16 @@ With some special chars: 🚀`,
 			},
 		},
 		"overwrites existing file atomically": {
-			file: File{
+			file: &File{
 				Path:    "test.txt",
 				Content: "Updated content",
 			},
 			setupFunc: func(tempDir string, fm *FileManager) error {
 				// Create initial file
-				initialFile := File{Path: "test.txt", Content: "Initial content"}
+				initialFile := &File{Path: "test.txt", Content: "Initial content"}
 				return fm.WriteFile(initialFile)
 			},
-			verifyFunc: func(t *testing.T, tempDir string, file File) {
+			verifyFunc: func(t *testing.T, tempDir string, file *File) {
 				// Verify the file was overwritten with new content
 				fullPath := filepath.Join(tempDir, file.Path)
 				content, err := os.ReadFile(fullPath)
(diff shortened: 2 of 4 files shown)


cli/internal/cmd/typer/typer.go
```diff
@@ -1,11 +1,14 @@
 package typer
```
fxenik, line 1: Let's skip the typer command for now

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/rudderlabs/rudder-iac/cli/internal/typer/generator/platforms/kotlin"
+	"github.com/rudderlabs/rudder-iac/cli/internal/typer/plan"
+)
+
+var (
+	orchestratorLog = logger.New("typer.orchestrator")
+)
+
+// Orchestrator coordinates the RudderTyper code generation process
+type Orchestrator struct {
```
fxenik, line 20: Let's actually rename this to `RudderTyper`, as mentioned in the design doc, to make it clear that this is the entry point for the typer logic

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+	deps app.Deps
```
fxenik, line 21: One thing to always keep in mind is that we should try to keep dependencies to components as minimal as possible. Never add something that's not needed because then it adds unnecessary coupling that you have to be take care of.

In this case, app.Deps is bringing in too much, like all the available Providers, that are required by other cli commands.

app.Deps is like a way for the cli to initialize common dependencies at the start. What we are doing in practice in other commands is that they load dependencies at the start and then initialize the components they need by passing only the dependencies that these components require. One way to think about it is that in almost any case, only cli commands need a dependency to app.Deps, while other components need to specify their own dependencies.

To start with, let's have only one dependency for RudderTyper, the PlanProvider. The way that's typically done in go is that that component with a dependency (in this case RudderTyper) defines an interface for that dependency (in this case PlanProvider interface). Whoever is going to instantiate RudderTyper will be responsible for passing an instance of that PlanProvider interface. The PlanProvider interface itself should expose just as much functions as the RudderTyper component requires. In a sense, its like the component dictates what it needs from the dependency.

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// GenerationOptions contains configuration for code generation
+type GenerationOptions struct {
+	TrackingPlanID string
```
fxenik, line 26: That tracking plan id needs to be an internal detail of the PlanProvider, according to the design doc, not something baked in RudderTyper

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+		"platform", options.Platform,
+		"outputPath", options.OutputPath)
+
+	// Step 1: Fetch tracking plan data
+	fmt.Println("(mock) 📥 Fetching tracking plan data...")
+	trackingPlan, err := o.fetchTrackingPlan(ctx, options.TrackingPlanID)
+	if err != nil {
+		return fmt.Errorf("fetching tracking plan: %w", err)
+	}
+	orchestratorLog.Debug("(mock) fetched tracking plan", "rules", len(trackingPlan.Rules))
```
fxenik, line 51: With a PlanProvider interface in place, this doesn't have to be mock, it can be a direct call to the PlanProvider's function

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+// generateCode generates platform-specific code from the tracking plan
+func (o *Orchestrator) generateCode(trackingPlan *plan.TrackingPlan, platform string) ([]*core.File, error) {
+	orchestratorLog.Debug("generating code for platform", "platform", platform, "rulesCount", len(trackingPlan.Rules))
+
+	switch platform {
+	case "kotlin":
+		return kotlin.Generate(trackingPlan)
+	default:
+		return nil, fmt.Errorf("unsupported platform: %s (supported platforms: kotlin)", platform)
+	}
```
fxenik, line 117: Consider implementing [that](https://www.notion.so/rudderstacks/RudderTyper-2-0-232f2b415dd0806799c4c8398bdb5653?source=copy_link#233f2b415dd0808a8599d43bb11cfb0f) component from the design doc, instead of hardcoding the strategy selection logic inside RudderTyper

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,150 @@
...
+	fileManager := core.NewFileManager(absOutputPath)
+
+	// Convert []*core.File to []core.File for WriteFiles
+	coreFiles := make([]core.File, len(files))
+	for i, file := range files {
+		coreFiles[i] = core.File{
+			Path:    file.Path,
+			Content: file.Content,
+		}
+		orchestratorLog.Debug("prepared file for writing", "path", file.Path)
```
fxenik, line 145: Let's stick to pointers as an interface for files, to avoid copying the file contents in memory. No need for a conversion here

fxenik (changes requested):

cli/internal/typer/generator/core/file_manager.go
```diff
@@ -47,21 +47,21 @@ func (fm *FileManager) WriteFile(file File) error {
 }
 
 // WriteFiles writes multiple files atomically as a batch operation
-func (fm *FileManager) WriteFiles(files []File) error {
+func (fm *FileManager) WriteFiles(files []*File) error {
```
fxenik, line 50: If we are changing this interface to pointers, let's make sure the other functions also have a pointer interface (e.g WriteFile or validate)

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,131 @@
+package typer
```
fxenik, line 1: Lets rename file to ruddertyper.go

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,131 @@
...
+		"platform", options.Platform,
+		"outputPath", options.OutputPath)
+
+	// Step 1: Fetch tracking plan data
+	fmt.Println("📥 Fetching tracking plan data...")
+	trackingPlan, err := rt.fetchTrackingPlan(ctx)
+	if err != nil {
+		return fmt.Errorf("fetching tracking plan: %w", err)
+	}
+	rudderTyperLog.Debug("fetched tracking plan", "rules", len(trackingPlan.Rules))
```
fxenik, line 57: Let's remove this since the same is reported by fetchTrackingPlan

cli/internal/typer/orchestrator.go
```diff
@@ -0,0 +1,131 @@
...
+	if len(files) == 0 {
+		rudderTyperLog.Debug("no files to write, returning early")
+		return nil
+	}
+
+	// Convert absolute path to avoid issues
+	absOutputPath, err := filepath.Abs(outputPath)
+	if err != nil {
+		return fmt.Errorf("resolving output path: %w", err)
+	}
```
fxenik, line 124: Let's remove this since for relative outputPath dirs file manager will anyway write to the current directory.

cli/internal/typer/orchestrator_test.go
```diff
@@ -0,0 +1,146 @@
...
+				OutputPath: "/tmp/out",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// TrackingPlanID is now handled by the PlanProvider, not in options
```
fxenik, line 81: Let's clean irrelevant LLM comments :)

cli/internal/typer/orchestrator_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			mockProvider := &mockPlanProvider{}
+			rudderTyper := NewRudderTyper(mockProvider)
+			if rudderTyper == nil {
+				t.Fatal("NewRudderTyper() returned nil")
+			}
+		})
+	}
+}
```
fxenik, line 52: Let's remove this one since it's just testing the constructor, and it is already implicitly tested elsewhere

cli/internal/typer/orchestrator_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func TestGenerationOptions(t *testing.T) {
```
fxenik, line 54: Let's remove that test since it's not testing anything for rudder typer

cli/internal/typer/orchestrator_test.go
```diff
@@ -0,0 +1,146 @@
...
+						return
+					}
+					if tt.wantErr != "" && err.Error() != tt.wantErr {
+						t.Errorf("Expected error '%s', got '%s'", tt.wantErr, err.Error())
+					}
+				} else {
+					if err != nil {
+						t.Errorf("Expected no error but got: %v", err)
+					}
+				}
```
fxenik, line 142: We can replace those assertions with calls from the assert package: 	github.com/stretchr/testify/assert

fxenik: Let's add also at least a test to verify that expected files have been created. Let's only verify filename and location for now since we don't really care about kotlin specifics.

fxenik (changes requested):

PR #188, opened 2025-09-25
feat: use new event stream sources APIs

shekhar-rudder:
This PR updates the event stream sources implementation to use the new API endpoints and response format.

**Changes**

  - API Endpoint: Changed from /v2/sources to /v2/event-stream-sources for create, update, and delete operations
  - Field Naming: Updated source_definition field to type with lowercase values (e.g., Javascript → javascript, ReactNative →
  react_native)

**Technical Details**

  - Updated all client methods (Create, Update, Delete) to work with the new response format
  - Updated test cases to reflect the new API endpoints and response structure
  - All source definition values are now lowercase and use snake_case where applicable

### DIFF
--- a/api/client/event-stream/source/source.go
+++ b/api/client/event-stream/source/source.go
@@ -9,7 +9,7 @@ import (
 	"github.com/rudderlabs/rudder-iac/api/client"
 )
 
-const prefix = "/v2/sources"
+const prefix = "/v2/event-stream-sources"
 
 type SourceStore interface {
 	Create(ctx context.Context, source *CreateSourceRequest) (*EventStreamSource, error)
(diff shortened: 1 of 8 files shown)


api/client/event-stream/source/source.go
```diff
@@ -73,7 +73,7 @@ func (r *rudderSourceStore) Delete(ctx context.Context, sourceId string) error {
 }
 
 func (r *rudderSourceStore) GetSources(ctx context.Context) ([]EventStreamSource, error) {
-	data, err := r.client.Do(ctx, "GET", prefix, nil)
+	data, err := r.client.Do(ctx, "GET", "/v2/sources", nil)
```
shekhar-rudder, line 76: Even `GetSources` should use the newer endpoint. But I'll update it in the tracking plan PR

am6010 (approved):

PR #461, opened 2026-03-13
feat(typer-kotlin): support custom annotation or compose stability via composeStable option

akashrpo:
🔒 Scanned for secrets using gitleaks 8.30.0

## Ticket
[DAW-3131](https://linear.app/rudderstack/issue/DAW-3131)

## Summary
Add `composeImmutable` option to the Kotlin code generator that applies `@androidx.compose.runtime.Immutable` to generated data classes (and concrete sealed subclasses) so Jetpack Compose treats them as stable, avoiding unnecessary recompositions.

## Changes
- New `composeImmutable` option in `KotlinOptions` (`--option composeImmutable=true`).
- When enabled, the generator adds `@Immutable` to every `data class` (top-level, nested, and concrete sealed subclasses) and emits `import androidx.compose.runtime.Immutable`. Sealed parents (abstract) and enums (already stable) are skipped.
- Generic `Annotations []string` on `KotlinDataClass` / `KotlinSealedSubclass` and `Imports []string` on `KotlinContext`, plumbed through `dataclass.tmpl`, `sealedclass.tmpl`, and `Main.kt.tmpl`. Keeps the door open for future annotations without more template churn.
- README updated under "Available Options" with the option description, runtime requirement (`androidx.compose.runtime` on the consuming module's classpath), and rationale for `@Immutable` over `@Stable`.

## Testing
- **Unit**: `TestGenerateWithComposeImmutable` reuses `testutils.GetReferenceTrackingPlan()` and snapshot-diffs against a committed `Main.kt`.
- **Compile validation**: the snapshot lives inside the validator Gradle project at `testdata/validator/.../composeimmutable/Main.kt`. `make typer-kotlin-validate` compiles it (alongside a stub `androidx.compose.runtime.Immutable` annotation in the validator project so we don't pull real Compose deps) and runs the existing test suite. Verified locally — all 12 tests pass, both Main.kt files compile.
- **Regenerate**: `make typer-kotlin-update-testdata` produces both vanilla and compose-immutable snapshots in one go.
- `make test` and `make lint` clean.

## Risk / Impact
**Low** — opt-in flag, default generator output unchanged. Adds compile coverage where there was none.

## Checklist
- [x] Ticket linked
- [x] Tests added/updated (unit + compile)
- [x] README updated
- [x] No breaking changes


cli/internal/typer/generator/platforms/kotlin/generator_test.go
```diff
@@ -81,3 +85,141 @@ func TestGenerateWithCustomOutputFileName(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		Metadata: plan.PlanMetadata{
+			TrackingPlanID:      "tp_123",
+			TrackingPlanVersion: 1,
+		},
+	}
+}
```
dhawal1248: **Consider reusing `testutils.GetReferenceTrackingPlan()` instead of this custom plan.**

The reference plan already covers data classes, nested classes, and sealed variants — exactly the shapes this feature annotates. Running it with `ComposeStable: true` and snapshotting a second expected file would:

- Mirror the `TestGenerate` pattern (one call site, one expected file).
- Keep the compose test in sync automatically as the reference plan evolves — right now any change to the reference plan leaves this bespoke plan drifting.
- Drop ~120 lines of hand-rolled plan setup.

The only cost is a bigger expected file, but that's already the precedent (`testdata/validator/.../Main.kt` is 1620 lines).

cli/internal/typer/generator/platforms/kotlin/testdata/compose_stable/compose_stable.kt
```diff
@@ -0,0 +1,282 @@
...
+import com.rudderstack.sdk.kotlin.core.Analytics
+import com.rudderstack.sdk.kotlin.core.internals.models.RudderOption
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
+import androidx.compose.runtime.Immutable
```
dhawal1248: **This file is string-diffed only — not compiled.**

The existing `testdata/validator/` runs the reference `Main.kt` through a real Kotlin compiler via `make typer-kotlin-validate` (Makefile:60-62). That's what catches issues like a bad import path. This new file bypasses that.

If you combine with recommendation #1 (reuse reference plan), add a Makefile target that regenerates a compose-stable variant into the validator project and compiles it. Otherwise a broken `androidx.compose.runtime.Immutable` import (or any future annotation addition) could regress without CI noticing.

cli/internal/typer/generator/platforms/kotlin/options.go
```diff
@@ -9,6 +9,7 @@ import (
 type KotlinOptions struct {
 	PackageName    string `mapstructure:"packageName" description:"Package name for generated Kotlin code (e.g., com.example.analytics)"`
 	OutputFileName string `mapstructure:"outputFileName" description:"Name of the generated Kotlin file (e.g., MyEvents.kt). Defaults to Main.kt"`
+	ComposeStable  bool   `mapstructure:"composeStable" description:"Add @Immutable annotations to generated data classes for Jetpack Compose stability"`
```
dhawal1248: **README isn't updated for this option.**

`cli/internal/typer/generator/platforms/kotlin/README.md` §"Available Options" (line 607) currently documents only `packageName`. Please add a `composeStable` entry covering:

- What it does (`@Immutable` on generated data classes + sealed subclasses).
- The `androidx.compose.runtime` runtime dependency requirement — without it on the classpath, the generated file will not compile. Users need to know this before flipping the flag.
- Why `@Immutable` and not `@Stable` (the rationale already in the PR description is worth preserving in-repo).

dhawal1248: **Naming: `composeStable=true` produces `@Immutable` — slightly surprising.**

The commit title ("support custom annotation or compose stability") also hints at broader intent that got narrowed down in implementation. Two ways to resolve:

1. Rename to `composeImmutable` — precise, matches what's actually emitted.
2. Keep `composeStable` but reshape as `composeStability=immutable|stable|none` so future `@Stable` support slots in without another flag.

Don't need to build the enum now, but picking a name that doesn't lock you in is worth a minute. Happy either way — just flagging before it ships and becomes an API.

cli/internal/typer/generator/platforms/kotlin/generator.go
```diff
@@ -620,6 +624,34 @@ func isEmptySchema(schema *plan.ObjectSchema) bool {
 	return schema != nil && len(schema.Properties) == 0
 }
 
+// applyComposeImmutable adds @Immutable to all data classes (top-level, nested, and sealed subclasses)
+// and the corresponding import. Sealed classes and enums are skipped — sealed classes are abstract
+// and enums are already treated as stable by the Compose compiler.
+func applyComposeImmutable(ctx *KotlinContext) {
+	annotations := []string{"Immutable"}
```
dhawal1248: Nit: `annotations := []string{"Immutable"}` is shared across every data class and sealed subclass (same underlying slice gets assigned to each `.Annotations` field). Nothing mutates it today, so this is safe — but if some future code path ever appends to a class's `Annotations`, it'll cross-contaminate every other class that shares the slice.

Either allocate per class (`dc.Annotations = []string{"Immutable"}`) or leave a comment noting the slice is treated as immutable.

dhawal1248 (commented):
A few suggestions to tighten this up — these would bring it in line with existing Kotlin-generator conventions.

dhawal1248 (changes requested):

cli/internal/typer/generator/platforms/kotlin/options.go
dhawal1248: We would also need to document this option in our public docs
cc @sivashanmukh

dhawal1248 (approved):

PR #514, opened 2026-04-16
fix(typer/swift): wrap multi-type union defaults in case constructors

dhawal1248:
## 🔗 Ticket

Resolves [DEX-317](https://linear.app/rudderstack/issue/DEX-317)

---

## Summary

The Swift generator emitted bare literals (`= true`, `= false`, `= "beta"`) as default values for variant-discriminator fields whose type is a multi-type union enum (e.g. `PropertyFeatureFlag` with `.bool(Bool)` / `.string(String)` cases). Those defaults do not type-check — Swift enums are not `ExpressibleByBooleanLiteral`/`ExpressibleByStringLiteral` — so the generated code would fail to compile. The golden-file test only did a string compare on the generator output, so the bug slipped through.

This fix detects the `.value`-suffixed serialize expression (the marker for multi-type enum fields) and wraps the discriminator's match value in the appropriate case constructor, producing `= .bool(true)`, `= .bool(false)`, `= .string("beta")`.

---

## Changes

- Added `multiTypeLiteral` helper in `generator.go` that maps a Go literal to the matching Swift case constructor (`.bool`, `.string`, `.int`, `.double`, `.null`).
- Extended the discriminator-default branch in `createVariantEnum` to call `multiTypeLiteral` when the property's serialize expression ends in `.value` (i.e. the field is a multi-type union enum).
- Regenerated the Swift testdata (`RudderTyper.swift`) so the three `CustomTypeFeatureConfig` cases now emit valid defaults.

---

## Testing

- `go test ./cli/internal/typer/generator/platforms/swift/... -count=1` — passes.
- Manually verified the generated defaults type-check against `PropertyFeatureFlag`.
- End-to-end compilation against the Swift SDK is covered by the upcoming validator work (DEX-319+) which surfaced this bug.

---

## Risk / Impact

Low.

Scope is limited to Swift code generation. The only behavioral change is in the default-value literal for multi-type union discriminator fields — the previous output did not compile, so there are no downstream consumers of the broken form.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/typer/generator/platforms/swift/generator.go
+++ b/cli/internal/typer/generator/platforms/swift/generator.go
@@ -615,6 +615,27 @@ func createMultiTypeEnum(name, comment string, types []plan.PropertyType) (*Swif
 	return &SwiftMultiTypeEnum{Name: name, Comment: comment, Cases: cases}, nil
 }
 
+// multiTypeLiteral formats a discriminator literal as a multi-type enum case
+// expression (e.g. `.bool(true)`, `.string("beta")`) so it type-checks against
+// the generated union enum.
+func multiTypeLiteral(value any) string {
+	if value == nil {
+		return ".null"
+	}
+	switch value.(type) {
+	case string:
+		return fmt.Sprintf(".string(%s)", FormatSwiftLiteral(value))
+	case int, int32, int64:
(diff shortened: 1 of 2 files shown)


SKannaniOS (approved):
LGTM

PR #516, opened 2026-04-16
feat(typer/swift): add Swift validator skeleton + identify smoke test

dhawal1248:
## 🔗 Ticket

Resolves [DEX-319](https://linear.app/rudderstack/issue/DEX-319)

Parent: [DEX-311](https://linear.app/rudderstack/issue/DEX-311) — Add e2e tests for RudderTyper Swift.
Depends on: [DEX-317](https://github.com/rudderlabs/rudder-iac/pull/514) — the generated Swift testdata needs the multi-type union default fix to compile.

---

## Summary

Introduces a SwiftPM-based validator under `cli/internal/typer/generator/platforms/swift/testdata/validator/` that compiles the generated `RudderTyper.swift` against the real `rudder-sdk-swift` (pinned to 1.2.1) and runs XCTest cases asserting the events the typer emits. This PR lands the skeleton plus the `identify` smoke test; follow-up PRs (DEX-320/321/322) add the remaining coverage.

---

## Changes

- `Package.swift` pins `rudder-sdk-swift` at `1.2.1`, declares a `RudderTyper` library target and a `RudderTyperTests` test target.
- `.gitignore` inside the validator excludes `.build/`, `.swiftpm/`, `Package.resolved`, and the copied-in `Sources/RudderTyper/*.swift` — only the test sources are checked in.
- `Sources/RudderTyper/.gitkeep` keeps the directory in git; the generated Swift is copied in by the Makefile target at test time.
- `Sources/RudderTyperTests/EventValidationPlugin.swift` intercepts `onProcess` events, drops them before the network layer, and exposes `validateCount` / `validateNext` helpers that compare events using canonical sorted-key JSON (sidesteps `NSDictionary` / `AnyCodable` deep-equality issues).
- `Sources/RudderTyperTests/RudderTyperTests.swift` sets up `Analytics` with `trackApplicationLifecycleEvents: false` + `SessionConfiguration(automaticSessionTracking: false)` (otherwise the SDK injects lifecycle and session events that pollute the plugin's event list) and asserts the `identify` flow.
- `Makefile`: new `typer-swift-validate` target that copies the generated testdata into the package and runs `swift test`.

---

## Testing

- The validator target is opt-in (`make typer-swift-validate`). It is not wired into CI in this PR — that is a follow-up task.
- Ran `make typer-swift-validate` locally against the DEX-317 branch: `identify` test passes.

[...]

PR #517, opened 2026-04-16
test(typer/swift): add group and screen validator coverage

dhawal1248:
## 🔗 Ticket

Resolves [DEX-320](https://linear.app/rudderstack/issue/DEX-320)

Parent: [DEX-311](https://linear.app/rudderstack/issue/DEX-311) — Add e2e tests for RudderTyper Swift.
Stacks on: [DEX-319 / PR #516](https://github.com/rudderlabs/rudder-iac/pull/516) — introduces the validator skeleton + `testIdentify`. Base branch is set to that PR so GitHub shows only the DEX-320 delta; merge DEX-319 first, then this PR retargets to `main` automatically.

---

## Summary

Adds `testGroup` and `testScreen` to the Swift validator, covering the `group` and `screen` entry points of the generated `RudderTyperAnalytics` against the real `rudder-sdk-swift`.

---

## Changes

- `testGroup` asserts that `typer.group(groupId:traits:)` dispatches through the SDK with the expected `groupId`. The generator routes group traits through `options.customContext` rather than onto `GroupEvent.traits`, so the assertion on `GroupEvent.traits` is an empty map — this is called out inline in a comment so a reader doesn't read it as a typo.
- `testScreen` covers two calls:
  - A full-shape call with nested `CustomTypeUserProfile`, a `category`, and explicit properties — asserts the generator serializes to the expected JSON shape including the injected `name` and `category` keys.
  - A minimal call (`screenName` + empty `ScreenProperties()`) — asserts the `name` field is injected by the typer even when the caller passes no properties.

---

## Testing

- `make typer-swift-validate` runs both new tests alongside the existing `testIdentify` (stacked on the DEX-319 validator skeleton + DEX-317 testdata fix): all three pass.
- No Go code or generator behaviour changes, so Go unit tests are unaffected.

---

## Risk / Impact

Low.

Test-only addition gated behind the opt-in `make typer-swift-validate` target. No production code paths touched; the Swift generator itself is unchanged.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
+++ b/cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
@@ -46,4 +46,54 @@ final class RudderTyperTests: XCTestCase {
             traits: ["active": false, "email": "jane.smith@example.com"]
         ))
     }
+
+    // MARK: - group
+
+    func testGroup() {
+        typer.group(
+            groupId: "company-xyz-789",
+            traits: GroupTraits(active: true)
+        )
+        analytics.flush()
+
+        validations.validateCount(1)
+        // Traits are routed through options.customContext, not onto GroupEvent.traits.
+        validations.validateNext(.group(groupId: "company-xyz-789", traits: [:]))
+    }
+
+    // MARK: - screen
+
+    func testScreen() {
+        typer.screen(
+            screenName: "Dashboard",
+            properties: ScreenProperties(
+                profile: CustomTypeUserProfile(
+                    email: "user@example.com",
+                    firstName: "Alice",
+                    lastName: "Johnson"
+                )
+            ),
+            category: "Main Navigation"
+        )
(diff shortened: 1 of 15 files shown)


cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
```diff
@@ -46,4 +46,54 @@ final class RudderTyperTests: XCTestCase {
...
+    func testGroup() {
+        typer.group(
+            groupId: "company-xyz-789",
+            traits: GroupTraits(active: true)
+        )
+        analytics.flush()
+
+        validations.validateCount(1)
+        // Traits are routed through options.customContext, not onto GroupEvent.traits.
+        validations.validateNext(.group(groupId: "company-xyz-789", traits: [:]))
```
akashrpo, line 61: Asserting traits: [:] passes even if the generator stops routing traits through options.customContext. Can we also assert active: true lands in the custom context?

akashrpo (approved):

SKannaniOS (approved):
LGTM

PR #519, opened 2026-04-16
test(typer/swift): add track edge-case validator coverage

dhawal1248:
## 🔗 Ticket

Resolves [DEX-322](https://linear.app/rudderstack/issue/DEX-322)

Parent: [DEX-311](https://linear.app/rudderstack/issue/DEX-311) — Add e2e tests for RudderTyper Swift.
Stacks on: [DEX-321 / PR #518](https://github.com/rudderlabs/rudder-iac/pull/518) — core track coverage. Base branch is set to that PR so GitHub shows only the DEX-322 delta; once the parents merge in order this retargets to `main` automatically.

Completes the `testIdentify` → `testGroup/Screen` → core-track → edge-track stack and closes out the DEX-311 initiative on the tests side.

---

## Summary

Adds six XCTest cases covering the generator's less-obvious track shapes — variant discriminators, primitive enum edges, and empty-struct serialization. Each test's header comment explicitly calls out the Swift-specific divergence from the Kotlin suite so a reader doesn't mistake the difference for a bug.

---

## Changes

- **`testTrackEventWithVariants`** — asserts the top-level variants discriminator routes the right payload case through the generated enum (the `Event With Variants` event's `.caseMobile` shape).
- **`testTrackUserSignedUpFeatureConfigs`** — exercises all four `feature_config` cases (`true` / `false` / `beta` / default). The Swift generator emits an empty `Default` struct, so the default case serializes to `{}`. The Kotlin suite's `Default` carries `featureFlag` — the divergence is called out inline.
- **`testTrackUserSignedUpEnums`** — asserts primitive enum serialization. `PropertyEnabled` and `PropertyMixedValue` are String-backed in Swift (rawValue `"true"` / `"2.5"`), so the serialized JSON carries strings rather than the booleans / numbers the Kotlin test asserts. Called out in the header comment.
- **`testTrackUserSignedUpUserAccess`** — `CustomTypeUserAccess.Default` has no fields in Swift, so `user_access` serializes to `{}` rather than Kotlin's `{"active": …}`.
- **`testTrackUserSignedUpAllEnums`** — combined-enums sanity check with every enum variant populated at once, guarding against cross-talk between the generated raw-value mappings.
- **`testTrackUserSignedUpUnitSerialization`** — empty-object property. Swift models `nestedEmptyObjectNoAdditionalProps` as `[String: Any]?` (Kotlin models it as `Unit`), so the test passes `[:]` and asserts the key is serialized as `{}`.

---

## Testing

- `make typer-swift-validate` runs all 12 tests (this set + identify + group + screen + the three core track cases) stacked on DEX-319 / DEX-320 / DEX-321 + DEX-317's testdata fix: all pass.
- No Go code or generator behaviour changes.

---

## Risk / Impact

Low.

Test-only addition gated behind the opt-in `make typer-swift-validate` target. No production code paths touched; the Swift generator itself is unchanged.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
+++ b/cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
@@ -255,4 +255,209 @@ final class RudderTyperTests: XCTestCase {
             "profile": ["email": "iot@example.com", "first_name": "George"],
         ]))
     }
+
+    // MARK: - track (variants discriminator)
+
+    func testTrackEventWithVariants() {
+        typer.trackEventWithVariants(
+            properties: .caseMobile(.init(
+                profile: CustomTypeUserProfile(
+                    email: "mobile.user@example.com",
+                    firstName: "Hannah",
+                    lastName: "Smith"
+                ),
+                tags: ["mobile", "app-user"]
+            ))
+        )
+        analytics.flush()
+
+        validations.validateCount(1)
+        validations.validateNext(.track(name: "Event With Variants", properties: [
+            "device_type": "mobile",
+            "profile": [
+                "email": "mobile.user@example.com",
+                "first_name": "Hannah",
+                "last_name": "Smith",
+            ],
+            "tags": ["mobile", "app-user"],
+        ]))
+    }
+
+    // MARK: - track (feature config variants)
+    //
+    // The Swift generator emits an empty `Default` struct (no fields) for the
+    // discriminator's fallback case. That is why the fourth "alpha" case here
+    // asserts an empty `feature_config` dict — differs from Kotlin where the
+    // Default case carries a `featureFlag` field.
+
+    func testTrackUserSignedUpFeatureConfigs() {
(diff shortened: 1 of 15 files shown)


cli/internal/typer/generator/platforms/swift/testdata/validator/Sources/RudderTyperTests/RudderTyperTests.swift
```diff
@@ -255,4 +255,209 @@ final class RudderTyperTests: XCTestCase {
...
+            "active": true,
+            "feature_config": [:] as [String: Any],
+            "profile": ["email": "feature.alpha@example.com", "first_name": "Alpha", "last_name": "User"],
+        ]))
+    }
+
+    // MARK: - track (primitive enums)
+    //
+    // Swift emits `PropertyEnabled` and `PropertyMixedValue` as String-backed
+    // enums (rawValue is a String like "true" / "2.5"), so the serialized JSON
```
akashrpo, line 345: Should enabled and mixed_value be serialized as strings ("true", "2.5") instead of real booleans/numbers? Is this intended?

akashrpo (approved):

PR #522, opened 2026-04-21
feat(migrate): preserve original YAML key and list order

dhawal1248:
## 🔗 Ticket

Resolves [DEX-328](https://linear.app/rudderstack/issue/DEX-328)

---

## Summary

Before this change, `rudder-cli migrate` rewrote spec files with keys alphabetized by the `yaml.v3` encoder. A version bump + a few ref rewrites landed in the diff alongside hundreds of lines of reorder noise, and users had to squint to find the real semantic changes. This PR makes the migrate write path emit YAML whose key and list-item order matches the user's original file — so the `git diff` after migrate reflects only the intentional migrations.

---

## Changes

- `specs.New()` now also parses the raw bytes into a `yaml.Node` and caches it on the `Spec` (accessible via `OriginalNode()`), so the migrator knows the original ordering.
- Added `formatter.YAMLOrderedFormatter`, which encodes the migrated spec and then reorders the result in place to match the cached original:
  - Mappings: keys restored to the original position; keys added by the migration are appended at the end of their parent; keys removed by the migration are dropped.
  - Sequences of objects: items matched across sides by an identity field, probed in order — `id`, `$ref`, `property` (all three schema-unique within their lists in rudder-iac specs). Sequences whose items carry none of those fall back to positional matching.
- `migrator.WriteSpecs` constructs a per-file `YAMLOrderedFormatter` seeded with that file's original node.
- The previous `YAMLFormatter` is untouched, so non-migrate write paths (apply, import, export) behave exactly as before.

---

## Testing

- Unit tests in `cli/internal/project/formatter/yaml_test.go`: key-order preservation on no-op, value-only change keeps key position, added key appended at end, removed key dropped, nested mapping order preserved, sequence matched by `id` / `$ref` / `property`, identity-key precedence (`id` wins over `$ref`/`property`), positional fallback when no identity key, scalar sequence preserved positionally, new list item without matching id appended, nil original leaves alphabetized output, duplicate identity on orig side preserves orig cardinality.
- Integration tests in `cli/internal/project/migrator/migrator_test.go`:
  - `TestWriteSpecs_PreservesOriginalOrder` — full `specs.New` → mutate → `WriteSpecs` on a valid rudder/v0.1 properties spec with nested propConfig blocks; also exercises the datacatalog-style `propConfig` → `config` rename (brand-new parent key appended at end of each property).
  - `TestWriteSpecs_TrackingPlan_PreservesOriginalOrder` — valid rudder/v0.1 tracking plan; reshuffles both the id-keyed rules list and the $ref-keyed inner properties list and asserts both are restored.
- Manual: ran the built binary against `cli/tests/testdata/project/create/` fixtures — the resulting `git diff` now shows only the version bump and semantic ref rewrites, no reorder churn.
- `make test` and `make lint` pass clean.

---

## Risk / Impact

Low.
Notes: changes are isolated to the migrate write path. The default `YAMLFormatter` used everywhere else is unchanged. Migration output may still differ stylistically from the original (comments and blank lines are lost — `yaml.v3` doesn't preserve them through a decode/encode cycle), but ordering is faithful.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/project/formatter/yaml_test.go
+++ b/cli/internal/project/formatter/yaml_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
 )
 
 func TestYAMLFormatter_Format(t *testing.T) {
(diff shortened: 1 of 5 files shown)


cli/internal/project/formatter/yaml_test.go
```diff
@@ -223,3 +224,316 @@ enabled: true
...
+				},
+			},
+			expected: heredoc.Doc(`
+				items:
+				  - id: "a"
+				    v: 99
+				  - id: "a"
+				    v: 99
+			`),
+		},
```
fxenik, line 420: Although that's not going to be a use case we'll see in practice, I feel the correct behavior would be to keep the cardinality of input, not original. Original should used only to influence the order of input, not change it's contents.

cli/internal/project/formatter/yaml_test.go
```diff
@@ -223,3 +224,316 @@ enabled: true
...
+				},
+			},
+			expected: heredoc.Doc(`
+				properties:
+				  - $ref: "#/properties/api_tracking/api_method"
+				    required: true
+				  - $ref: "#/properties/api_tracking/username"
+				    required: false
+			`),
+		},
```
fxenik, line 375: In practice, our migration is going to convert `$ref` keys to `property` keys (plus changing the reference string format. The ideal scenario would be to preserve ordering even if original has `$ref` keys and input has `property`, right?

cli/internal/project/formatter/yaml_test.go
```diff
@@ -223,3 +224,316 @@ enabled: true
...
+					map[string]any{"foo": "1"},
+					map[string]any{"foo": "2"},
+				},
+			},
+			expected: heredoc.Doc(`
+				items:
+				  - foo: "1"
+				  - foo: "2"
+			`),
+		},
```
fxenik, line 462: I'm not sure I get what this test is validating, it looks like both original and input are the same. In this scenario though, wouldn't it be more appropriate to just keep the input without further processing?

fxenik (approved):

PR #545, opened 2026-04-28
feat(typer/typescript): add generator skeleton with identify event

dhawal1248:
## 🔗 Ticket

Resolves [DEX-331](https://linear.app/rudderstack/issue/DEX-331/add-generator-skeleton-with-identify-event)

---

## Summary

Lands the v1 TypeScript typer platform, mirroring the role of the Swift skeleton PR (#516) but trimmed to identify-only so the cross-cutting TS design decisions can ship in a single reviewable change. Subsequent tickets layer in `track`, `screen`, `group`, custom-type aliases, multi-type unions, variants, and the validator project.

---

## Changes

- New platform package `cli/internal/typer/generator/platforms/typescript/` implementing `core.Generator`. File layout, naming-helper conventions, template wiring, section dividers, and collision-handler pattern all mirror Swift 1:1.
- Naming module: `PascalCase` for types, `camelCase` for properties, reserved + contextual TS keywords renamed via `_` suffix, identifier validity check for non-Latin / kebab / dotted JSON keys (emitted as quoted properties to keep on-the-wire serialization exact).
- Type mapping: `string` / `integer|number → number` / `boolean` / `null` / arrays (`T[]` or `Array<A | B>` for unions) / objects → `Record<string, unknown>`. Multi-type unions dedup so `integer | number` collapses to `number`.
- Templates: `RudderTyper.ts.tmpl`, `disclaimer.tmpl`, `interface.tmpl`, `ruddertyper.tmpl`. Single-file ESM emit; `RudderTyper` class with constructor-injected `RudderAnalytics` from `@rudderstack/analytics-js`; generator does not call `load()`.
- Registered the platform in `cli/internal/typer/generator/platforms.go` and the CLI allow-list (`cmd/typer/typer.go`, `cmd/typer/options.go`).
- `testutils/generate_reference_plan.go` regen binary, `testdata/RudderTyper.ts` golden snapshot, and `make typer-typescript-update-testdata` target (mirrors Swift/Kotlin).
- Non-identify event types are accepted by the generator but skipped with a warning so the shared reference tracking plan keeps validating across all platforms.

### Design decisions baked in (per ticket)

- Class (matches Kotlin/Swift; accepts the tree-shaking trade-off for consistency)
- Module format: ESM
- Single-file emit
- Optional properties: `prop?: T` (not `T | undefined`)
- Null handling: explicit `T | null`, distinct from optional
- Open objects: `Record<string, unknown>`
- JSDoc on every generated symbol
- Generated file compiles under `strict: true` + `exactOptionalPropertyTypes: true`
- SDK injection via constructor; `RudderAnalytics` and `ApiOptions` imported as types from `@rudderstack/analytics-js`
- Inline vs named: v1 uses named interfaces only for event traits (single-use); custom-type aliases deferred to follow-up tickets

---

## Testing

- **Unit tests**: `naming_test.go` (PascalCase / camelCase / reserved-word renaming / tokenization edge cases), `escape_test.go` (string-literal escaping, JSDoc comment escaping including `*/` and newline collapse), `typemapping_test.go` (primitives, arrays, multi-type unions including `integer | number` dedup, custom-type fallback, identifier-validity check).
- **Golden file**: `generator_test.go` runs the full pipeline against the shared reference plan and diffs against `testdata/RudderTyper.ts`. `make typer-typescript-update-testdata` regenerates the snapshot.
- **Lint + tests**: `make lint` and `make test` both green; package coverage 83.9%.
- **Manual**: `rudder-cli typer options --platform=typescript` lists the option; `rudder-cli typer generate --platform=typescript ...` is in the allow-list.
- **Strict-tsc validation of the generated file**: still TODO per ticket ("verify manually") — recommend the reviewer paste the golden into a `tsc --strict --exactOptionalPropertyTypes` project before approving.

---

## Risk / Impact

Low.

- Net-new platform; no behavior change to existing Kotlin / Swift typers.
- Shared `core` types and `plan` types are read-only from this package.
- Reference tracking plan in `testutils/` was not modified (would have rippled into Swift / Kotlin golden files).

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated
- [x] No breaking changes


cli/internal/typer/generator/platforms/typescript/generator.go
```diff
@@ -0,0 +1,499 @@
...
+	if !isEmptySchema(&rule.Schema) {
+		interfaceName, err := getOrRegisterEventInterfaceName(rule, nr)
+		if err != nil {
+			return nil, err
+		}
+		method.MethodArguments = append(method.MethodArguments, TSMethodArgument{
+			Name:     "traits",
+			Type:     interfaceName,
+			Comment:  "The traits to include with this event",
+			Optional: true,
```
akashrpo, line 487: `Optional: true` is hardcoded on the `traits` argument regardless of plan state. If the schema has any required trait, the caller can still skip the entire object , the interior interface enforces required fields at the property level, but an optional parameter bypasses that.

```ts
rudderTyper.identify("user_1");  // compiles even when `email` is required by plan

File: cli/internal/typer/generator/platforms/typescript/generator.go [hunk 2]
```diff
@@ -0,0 +1,499 @@
...
+
+// ========== Analytics Method Builder ==========
+
+func buildIdentifyMethod(rule *plan.EventRule, nr *core.NameRegistry) (*TSAnalyticsMethod, error) {
+	methodName, err := getOrRegisterEventMethodName(rule, nr)
+	if err != nil {
+		return nil, err
+	}
+
+	method := &TSAnalyticsMethod{
```
akashrpo, line 461: The [JS SDK supported-API docs](https://www.rudderstack.com/docs/sources/event-streams/sdks/rudderstack-javascript-sdk/supported-api/#identify) list five valid identify overloads:


rudderanalytics.identify(userId, [traits], [apiOptions], [callback]);
rudderanalytics.identify(userId, [traits], [callback]);
rudderanalytics.identify(userId, [callback]);
rudderanalytics.identify(traits, [apiOptions], [callback]);
rudderanalytics.identify(traits, [callback]);
The generator emits one shape with required userId: string and no callback arg. Anonymous identify and callback-based observability — both real call patterns — become impossible through the typer; users have to drop back to raw analytics.identify to use them.

cli/internal/typer/generator/platforms/typescript/generator.go
```diff
@@ -0,0 +1,499 @@
...
+
+// resolvePropertyType returns the TS type expression for a property in an
+// interface. Multi-type unions (including `T | null` for nullable properties)
+// are emitted inline so the IR's distinction between optional (`prop?: T`) and
+// nullable (`T | null`) survives into the output.
+//
+// Custom types currently resolve to their underlying primitive — named type
+// aliases for custom types are intentionally deferred to follow-up tickets so
+// the identify-only landing stays small.
+func resolvePropertyType(prop *plan.Property) (string, error) {
```
akashrpo, line 194: `plan.Property.Enum` (plan.go:48) is populated by the spec loader and consumed by the Swift platform (swift/context.go — `SwiftEnum/SwiftEnumValue`). The TS path never reads it.

Effect: `tier: { type: string, enum: ["free", "pro"] }` renders as `tier: string` instead of `tier: "free" | "pro"`. Loses the literal-narrowing the typer is supposed to provide.

cli/internal/typer/generator/platforms/typescript/generator.go
```diff
@@ -0,0 +1,499 @@
...
+func resolveCustomType(ct *plan.CustomType) (string, error) {
+	if ct.Type == plan.PrimitiveTypeArray {
+		if ct.ItemType == nil {
+			return "unknown[]", nil
+		}
+		return resolveArrayItem(ct.ItemType)
+	}
+	if ct.Type == plan.PrimitiveTypeObject {
+		// Open object — no per-custom-type interface generated yet.
+		return "Record<string, unknown>", nil
```
akashrpo, line 266: Inline-object traits with a populated `Schema` always resolve to `Record<string, unknown>`. Trait `address: { street: string, city: string }` loses field-level checking, which defeats the typer's main purpose.

cli/internal/typer/generator/platforms/typescript/generator.go
```diff
@@ -0,0 +1,499 @@
...
+		)
+		switch {
+		case plan.IsCustomType(t):
+			ct := plan.AsCustomType(t)
+			if ct == nil {
+				part = "unknown"
+			} else {
+				part, err = resolveCustomType(ct)
+			}
+		case plan.IsPrimitiveType(t):
```
akashrpo, line 241: When a trait is declared as a nullable typed array — i.e. `Types: [Array, Null]` with `ItemTypes: [String]` — this branch hits `mapPrimitiveToTSType(Array)`, which returns the open `unknown[]` and drops `ItemTypes` on the floor.

So the property renders as:

```ts
tags: unknown[] | null
```

instead of:

```ts
tags: string[] | null
```

Worth covering in `TestResolvePropertyType_MultiType` once resolved.

cli/internal/typer/generator/platforms/typescript/reserved.go
```diff
@@ -0,0 +1,89 @@
...
+	"as":         true,
+	"implements": true,
+	"interface":  true,
+	"let":        true,
+	"package":    true,
+	"private":    true,
+	"protected":  true,
+	"public":     true,
+	"static":     true,
+	"yield":      true,
```
akashrpo, line 63: The reserved-word table here drives whether a property name gets renamed (with `_` suffix) to avoid clashing with a JS/TS keyword. `eval` and `arguments` aren't in the list, so the generator treats them as ordinary identifiers.

Problem: under `"use strict"` (which the generated file runs in via `strict: true` in tsconfig), neither name is allowed as a variable, parameter, or property identifier — using them is a `SyntaxError`.

If a plan declares a trait literally named `eval`:

```yaml
traits:
  eval: { type: string }
```

we currently emit:

```ts
export interface IdentifyTraits {
    eval: string;  // SyntaxError under strict mode
}
```

Both `eval` and `arguments` should be in the reserved list so they get renamed (e.g. `eval_`, `arguments_`) like the other contextual keywords already are.

akashrpo (approved):
Rest LGTM. 
cc @fxenik

fxenik (approved):

PR #558, opened 2026-05-11
chore(ci): workflow cleanup (SEC-284)

aris1009:
Disables Go module cache restore in release workflow.

Ref: SEC-284


lokey (approved):
Approved: SEC-284 workflow-only changes

PR #559, opened 2026-05-11
test(typer/typescript): add core track validator coverage (DEX-340)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-340](https://linear.app/rudderstack/issue/DEX-340/add-core-track-validator-coverage)
Also covers [DEX-333](https://linear.app/rudderstack/issue/DEX-333/support-multi-type-property-unions) — multi-type property unions (the generator support landed with DEX-332; this PR adds the explicit unit-test coverage the ticket called for).

Stacks on: [DEX-332](https://linear.app/rudderstack/issue/DEX-332) — track event generation. Base branch is set to that PR so GitHub shows only the delta; once the parent merges, this retargets to `main` automatically.

---

## Summary

Adds 10 vitest cases covering every `track()` method the TypeScript generator emits from the reference plan, asserting on the analytics payloads captured by the existing MSW-based interceptor. Mirrors the structure of the Swift core-track suite (PR #518) and extends it to also smoke-test each remaining track method, since the TS surface area is small enough to enumerate exhaustively.

Also includes a small generator fix uncovered while writing the closed-empty-event test, and two unit tests that lock in the multi-type union design decisions for DEX-333.

---

### Track validator coverage (DEX-340)

- **`trackUserSignedUp` comprehensive** — string / number / boolean / array / nested object / string-literal enum / many optional-present fields in one call. Asserts the full `event.properties` shape verbatim.
- **`trackUserSignedUp` minimal** — only the required `active` + `profile` fields; confirms optionals are omitted (not serialized as `null`).
- **`trackUserSignedUp` device types** — one call per `PropertyDeviceType` literal (`mobile`, `tablet`, `desktop`, `smartTV`, `IoT-Device`); locks in the raw-value preservation including non-standard casing.
- **`trackUserSignedUp` hoisted / variant-base** — exercises the deepest-first hoisted `UserSignedUpContext.nestedContext` interface, arrays of `CustomTypeAddressDetails` / `CustomTypeUserProfile`, the variant-base custom types (`featureConfig`, `userAccess`), and every property-enum kind (`status` / `priority` / `rating` / `enabled`) in a single call.
- **`trackProductPremiumClicked`** — event name with embedded quotes plus `PropertyStatusCode` literal.
- **`trackVariableString`** — `$`-tokenized event name plus `PropertyDollarField` literal.
- **`trackEventWithNameCamelCase`** + **`trackEventWithNameCamelCase1`** — sanitization-collision counterparts; locks in that both event names round-trip on the wire unchanged.
- **`trackEmptyEventNoAdditionalProps`** — closed empty schema; no-arg call.
- **`trackEmptyEventWithAdditionalProps`** — open empty schema; arbitrary props pass through verbatim.
- Bumps `queueOptions.maxItems` from 1 to 100 in this file. The original value throttled the four rapid track calls in the device-types case (only one event reached MSW within 5s). 100 is closer to a realistic queue size and lets all four flush. `identify.test.ts` left alone since each of its tests dispatches a single event.

### Generator fix — closed-empty track events emit `{}` (d360b814)

- For events with `additionalProperties: false` and no declared properties, the generated `track()` previously passed `undefined` as the properties argument. Now passes `{}` so the wire payload is `properties: {}` regardless of how the SDK serializes a missing argument. Aligns with the canonical track-call template in `ts-spec.md`.
- Touches `generator.go`, regenerates `testdata/RudderTyper.ts`, and updates the unit-test assertion in `track_test.go`.

### Multi-type union test coverage (DEX-333)

- `TestBuildInterfaceWithNested_OptionalMultiType` — locks in `prop?: A | B` rendering (never `A | B | undefined`); `null` belongs in the union, `undefined` does not. Matters under `exactOptionalPropertyTypes`.
- `TestBuildInterfaceWithNested_MultiTypeInsideNestedObject` — a multi-type property inside a hoisted nested-object schema must resolve to a union on the nested interface, not collapse to `unknown`.
- Whole-struct comparisons per `CLAUDE.md` convention.

---

## Testing

- `make typer-typescript-validate` from the repo root: 12 tests pass (2 identify + 10 track). The Makefile target also runs `tsc --noEmit` against the generated `RudderTyper.ts`.
- `go test ./cli/internal/typer/generator/platforms/typescript/...` — all generator unit tests pass, including the new multi-type tests and the updated closed-empty assertion.

---

## Risk / Impact

Low.

The generator change is scoped to one branch of `buildTrackMethod` (closed-empty schema, no `additionalProperties`); previously emitted `undefined`, now emits `{}`. Callers that depend on the SDK's `undefined` handling are unaffected at the type level — only the literal value in the generated source changes. Test-only otherwise.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/typer/generator/platforms/typescript/testdata/validator/src/__tests__/track.test.ts
```diff
@@ -0,0 +1,349 @@
+import { RudderAnalytics } from "@rudderstack/analytics-js/bundled";
```
akashrpo, line 1: No context.ruddertyper assertion on any track event

cli/internal/typer/generator/platforms/typescript/testdata/validator/src/__tests__/track.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    expect(event.type).toBe("track");
+    expect(event.event).toBe("$Variable$String");
+    expect(event.properties).toEqual({ dollarField: "$variable_name" });
+  });
+
+  // ---- trackEventWithNameCamelCase ($eventWithNameCamelCase$!) ----
+
+  it("trackEventWithNameCamelCase dispatches the punctuated event name with optional field absent", async () => {
+    typer.trackEventWithNameCamelCase({});
```
akashrpo, line 297: This event's schema has only optional fields, but the generator marks `props` as required (`Optional: false` hardcoded in `buildTrackMethod`), so the caller is forced to pass `{}`. The test works around the bug instead of catching it.

akashrpo (approved):

PR #562, opened 2026-05-14
feat(typer/typescript): add support for variants (DEX-335)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-335](https://linear.app/rudderstack/issue/DEX-335/add-variant-discriminated-union-generation)

---

## Summary

Add TypeScript discriminated union generation for variant types in RudderTyper. The plan IR's `Variant` / `Variants` (discriminator-based if-then-else schemas) now emit idiomatic TypeScript discriminated unions instead of falling back to base-schema-only interfaces or being skipped entirely.

---

## Changes

- **`variants.go`** (new): Core variant processing — builds one interface per match value with a literal-typed discriminator, a default interface with the original type, and a union type alias joining all cases. Handles custom type variants and event rule variants.
- **`context.go`**: Added `TSVariantGroup` struct and `VariantTypes` field to `TSContext`.
- **`generator.go`**: Replaced variant warnings/skip with calls to variant processors. Simplified `resolveCustomTypeReference` (no more variant fallback). Updated stale comment.
- **`RudderTyper.ts.tmpl`**: Added "Variant Types" section between Custom Types and Nested Object Types.
- **`variants_test.go`** (new): Tests covering string, boolean, multi-type, and enum discriminators, multiple match values, missing default schemas, required-status merging, missing discriminator silent-skip.
- **`track_test.go`**: Updated existing tests from skip/base-only assertions to proper variant emission assertions.
- **`testdata/RudderTyper.ts`**: Golden file updated with all four variant groups.

### Variant shapes covered

| Discriminator type | Example | Reference plan type |
|---|---|---|
| String | `pageType: "search"` | `page_context` |
| Boolean | `active: true` | `user_access` |
| Multi-type (boolean \| string) | `featureFlag: "beta"` / `featureFlag: true` | `feature_config` |
| Enum | `deviceType: "mobile"` | `Event With Variants` |

### Design decisions

- **Discriminator collision**: If a variant case's payload has a property with the same name as the discriminator, the merge uses OR on required status (base wins on type, case can upgrade optional → required).
- **Missing-discriminator fallback**: Silently accepted (matches Kotlin's behavior) — the discriminator property just won't get literal narrowing.
- **Render order**: Union alternatives sorted alphabetically for deterministic output.
- **Default case**: Always emitted (matches Kotlin and Swift), even when cases are exhaustive.

---

## Testing

- Unit tests for all variant shapes (string/boolean/multi-type/enum discriminators)
- Golden file test (`TestGenerate`) exercises all four reference plan variant types end-to-end
- Edge cases: multiple match values per case, missing default schema, required-status merging, missing discriminator, enum discriminator in default
- `make lint` passes with 0 issues

---

## Risk / Impact

Low
- No changes to non-variant code paths (existing events, custom types, properties unaffected)
- Variant event rules that were previously skipped now emit methods (additive)
- Variant custom types that were previously base-schema-only now emit unions (type changes but property references resolve to the same name)

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/typer/generator/platforms/typescript/variants.go
```diff
@@ -0,0 +1,235 @@
...
+		return v
+	case bool:
+		if v {
+			return "True"
+		}
+		return "False"
+	default:
+		return fmt.Sprintf("%v", v)
+	}
+}
```
akashrpo, line 235: The `Default` case is meant as a fallback for values no named case handles, but right now its discriminator is the full enum — same as the named cases.

So `{ deviceType: "mobile" }` matches both `CaseMobile` and `Default`. When you do `if (e.deviceType === "mobile")`, TS can't narrow to just `CaseMobile` — it stays as `CaseMobile | Default`, and you lose access to case-specific fields like `tags`.

Restricting the Default's discriminator to values the named cases don't cover (e.g. `Exclude<PropertyDeviceType, "mobile" | "desktop">`) would fix it. For closed enums that's `never`, which is the right semantic — Default becomes unreachable.

akashrpo (approved):

PR #583, opened 2026-05-25
fix: use v1 item_type/item_types for tracking plan additionalProperties default

dhawal1248:
## Summary

  Fix tracking plan property `additionalProperties` defaulting for v1 array properties by using the migrated `item_type` / `item_types` fields instead of the removed
  `config["item_types"]` path.

  ## Problem

  As part of the v1 migration, array item types moved from property config to top-level spec fields. The tracking plan expansion/state path still tried to read
  `item_types` from `prop.Config`, and the expanded tracking plan property model was not preserving `ItemType` / `ItemTypes`.

  This caused incorrect defaulting of property-level `additionalProperties` in tracking plans, especially for array properties whose item types include `object`.

  ## Changes

  - added `ItemType` and `ItemTypes` to expanded tracking plan properties
  - preserved those fields during tracking plan ref expansion
  - updated tracking plan property defaulting to read `item_type` / `item_types` from the v1 fields
  - added coverage in the existing e2e project apply flow for an array property with `item_types: [object]`, asserting that the resulting tracking plan property gets
  `additionalProperties: true`

  ## Testing

  - `go test ./cli/internal/providers/datacatalog/state ./cli/internal/providers/datacatalog/localcatalog`
  - `go run ./cli/cmd/rudder-cli validate -l cli/tests/testdata/project/create`
  - `go run ./cli/cmd/rudder-cli validate -l cli/tests/testdata/project/update`

  Note: the full existing `TestProjectApply` flow could not be completed in this environment because the upstream API-backed test setup returned `401 Unauthorized`.


cli/internal/providers/datacatalog/state/trackingplan.go
```diff
@@ -388,6 +389,23 @@ func (args *TrackingPlanPropertyArgs) propertyByID(propertyID string) *TrackingP
...
+	switch {
+	case prop.ItemType != "":
+		return []string{prop.ItemType}
+	case len(prop.ItemTypes) > 0:
+		return prop.ItemTypes
+	}
+	return nil
+}
+
+func tpEventPropertyTypes(prop *localcatalog.TPEventProperty) []string {
```
abhimanyubabbar:
```suggestion
func getPropertyTypes(prop *localcatalog.TPEventProperty) []string {
```

cli/internal/providers/datacatalog/state/trackingplan.go
```diff
@@ -399,8 +417,9 @@ func getAdditionalPropertiesDefaultVal(prop *localcatalog.TPEventProperty) bool
 	if strings.HasPrefix(prop.Type, "#custom-type:") {
 		return false
 	}
-	hasObject := strings.Contains(prop.Type, "object")
-	hasArray := strings.Contains(prop.Type, "array")
+	types := tpEventPropertyTypes(prop)
+	hasObject := slices.Contains(types, "object")
+	hasArray := slices.Contains(types, "array")
```
abhimanyubabbar: Can we write a simple test case for the getAdditionalPropertiesDefaultVal as this is an important function and couldn't find a dedicated test for it

abhimanyubabbar (commented):
2 small nits

abhimanyubabbar (approved):
LGTM

PR #585, opened 2026-05-25
feat: add varsubst package foundation

dhawal1248:
## 🔗 Ticket

Resolves DEX-358

---

## Summary

Create the varsubst package foundation: the Resolver interface that all variable sources implement, the shared error/sentinel types, and the first concrete resolver — EnvResolver — which resolves variables from environment variables with a configurable prefix.

---

### Description

This ticket establishes the core abstractions for the variable substitution system and delivers the highest-priority resolver (environment variables).

### Resolver Interface

Define a `Resolver` interface that resolves a variable name to its string value:

```go
type Resolver interface {
    Resolve(name string) (value string, found bool)
}
```

The two-return-value pattern (`value` + `found bool`) lets callers distinguish between `resolved to empty string` and `not found` — important because an empty env var is a valid, intentional value.

### Error Types

Define the sentinel errors and the structured `SubstitutionError` type that will be used across the package:

```go
type SubstitutionError struct {
    Name     string   // variable name, e.g. "DB_PASSWORD"
    Line     int      // 1-indexed line number
    Column   int      // 1-indexed column number
    LineText string   // full text of the line for display
    Err      error    // underlying error (one of the sentinels)
}
```

Sentinel errors:

- `ErrUndefinedVariable` — `{{ .VAR }}` with no default, not found in any resolver
- `ErrInvalidVarSyntax` — malformed token: `{{ .123BAD }}`, unclosed `{{ .FOO`
- `ErrVarFileNotFound` — `--var-file` path does not exist
- `ErrVarFileParseFailed` — var file has invalid YAML or nested structures

### EnvResolver

Implements `Resolver`. At construction time, eagerly loads all environment variables matching a given prefix (default: `RUDDER_`), strips the prefix, and stores them in an in-memory map. `Resolve` is a simple map lookup.

Example: `RUDDER_DB_HOST=db.example.com` → stored as `DB_HOST` → `{{ .DB_HOST }}` resolves to `"db.example.com"`

```go
type EnvResolver struct {
    vars   map[string]string
    prefix string
}

func NewEnvResolver(prefix string) *EnvResolver
func (r *EnvResolver) Resolve(name string) (string, bool)
```

### Acceptance Criteria

- `varsubst` package exists with `resolver.go`, `errors.go`, `env_resolver.go`
- `Resolver` interface is defined
- All sentinel errors and `SubstitutionError` struct are defined
- `EnvResolver` loads env vars with prefix stripping
- Unit tests pass with full coverage of `EnvResolver`

---

## Testing

Table-driven unit tests for `EnvResolver` covering:

- Prefix stripping works correctly
- Case sensitivity (`RUDDER_db_host` vs `RUDDER_DB_HOST`)
- Empty value env var (`RUDDER_HOST=""` → resolves to `""`, `found=true`)
- Missing variable (`returns "", found=false`)
- Env vars without the matching prefix are ignored
- Multiple env vars loaded correctly

---

## Risk / Impact

Low

Notes (if any):

- This PR only adds the varsubst package foundation and corresponding unit tests.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/varsubst/resolver/env_resolver.go
+++ b/cli/internal/varsubst/resolver/env_resolver.go
@@ -23,6 +23,9 @@ func newEnvResolverFromEnviron(environ []string) Resolver {
 			continue
 		}
 		if strings.HasPrefix(key, defaultEnvPrefix) {
+			// Last-write-wins: if the same RUDDER_-prefixed key appears more than
+			// once in environ, the later occurrence silently overrides the earlier
+			// one. This mirrors the OS behaviour on lookup.
 			vars[strings.TrimPrefix(key, defaultEnvPrefix)] = value
 		}
 	}
--- a/cli/internal/varsubst/resolver/env_resolver.go
+++ b/cli/internal/varsubst/resolver/env_resolver.go
@@ -1,0 +1,36 @@
+package resolver
+
+import (
+	"os"
+	"strings"
+)
+
+const defaultEnvPrefix = "RUDDER_"
+
+type envResolver struct {
+	vars map[string]string
+}
+
+func NewEnvResolver() Resolver {
+	return newEnvResolverFromEnviron(os.Environ())
+}
+
+func newEnvResolverFromEnviron(environ []string) Resolver {
+	vars := make(map[string]string)
+	for _, env := range environ {
+		key, value, ok := strings.Cut(env, "=")
+		if !ok {
+			continue
+		}
+		if strings.HasPrefix(key, defaultEnvPrefix) {
+			vars[strings.TrimPrefix(key, defaultEnvPrefix)] = value
+		}
+	}
+
+	return &envResolver{vars: vars}
+}
+
+func (r *envResolver) Resolve(name string) (string, bool) {
+	value, found := r.vars[name]
+	return value, found
+}
--- a/cli/internal/varsubst/resolver/env_resolver_test.go
+++ b/cli/internal/varsubst/resolver/env_resolver_test.go
@@ -89,13 +89,39 @@ func TestNewEnvResolverFromEnviron(t *testing.T) {
 	}
 }
 
-func TestNewEnvResolver_LoadsRudderPrefixedVariables(t *testing.T) {
-	t.Setenv("RUDDER_ABC", "hello")
-	t.Setenv("ABC", "ignored")
+func TestNewEnvResolver(t *testing.T) {
+	t.Setenv("RUDDER_DB_HOST", "db.example.com")
+	t.Setenv("RUDDER_DB_PORT", "5432")
+	t.Setenv("RUDDER_EMPTY", "")
+	t.Setenv("DB_HOST", "ignored")
 
 	r := NewEnvResolver()
-	value, found := r.Resolve("ABC")
 
-	assert.Equal(t, "hello", value)
-	assert.True(t, found)
+	t.Run("strips prefix and loads value", func(t *testing.T) {
+		value, found := r.Resolve("DB_HOST")
+		assert.Equal(t, "db.example.com", value)
+		assert.True(t, found)
+	})
+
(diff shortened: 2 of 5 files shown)


cli/internal/varsubst/resolver/env_resolver.go
```diff
@@ -0,0 +1,36 @@
...
+func newEnvResolverFromEnviron(environ []string) Resolver {
+	vars := make(map[string]string)
+	for _, env := range environ {
+		key, value, ok := strings.Cut(env, "=")
+		if !ok {
+			continue
+		}
+		if strings.HasPrefix(key, defaultEnvPrefix) {
+			vars[strings.TrimPrefix(key, defaultEnvPrefix)] = value
+		}
```
abhimanyubabbar, line 30: As per this we would silently override the env variables to pick the last one right ?

cli/internal/varsubst/resolver/env_resolver.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+const defaultEnvPrefix = "RUDDER_"
+
+type envResolver struct {
+	vars map[string]string
+}
+
+func NewEnvResolver() Resolver {
+	return newEnvResolverFromEnviron(os.Environ())
```
abhimanyubabbar, line 15: It's better to inject the os.Environ() in the NewEnvResolver(envs []string) Resolver ?

Would allow us to test the envResolver in a better and easier way

cli/internal/varsubst/resolver/env_resolver_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := newEnvResolverFromEnviron(tt.environ)
+			value, found := r.Resolve(tt.lookup)
+			assert.Equal(t, tt.wantValue, value)
+			assert.Equal(t, tt.wantFound, found)
+		})
+	}
```
abhimanyubabbar, line 89: Given it's a small package, we can easily test via the constructor function which would be entry to managing the env's

cli/internal/varsubst/resolver/env_resolver.go
dhawal1248: yes

cli/internal/varsubst/resolver/env_resolver.go
dhawal1248: that is what the newEnvResolverFromEnviron function does..we use newEnvResolverFromEnviron to test and the actual could just call NewEnvResolver
IMHO, the logic of how to load the envs should be inside the envResolver.The caller does not need to care about how the envs are loaded

abhimanyubabbar: The logic to parse the env's is still there it's just that you get list of env's to work upon from outside. 

The envResolver should encapsulate only the part to loop over the env's and extract the ones based on RUDDER_ This logic is internal to EnvResolver. It also makes the test more deterministic to understand what will happen if we feed different types of env's. 

The caller can then simply provide you with all the data set for the env's to boot it up.

cli/internal/varsubst/resolver/env_resolver.go
abhimanyubabbar: Cool, better to write a comment there that it is acceptable to simply provide conflicting env's and system would pick the latest value.

dhawal1248: will do

cli/internal/varsubst/resolver/env_resolver.go
dhawal1248: will do: use t.SetEnv, set some env vars and test the public func NewEnvResolver()

abhimanyubabbar (approved):

PR #589, opened 2026-05-26
feat: add FileResolver for varsubst package

dhawal1248:
## 🔗 Ticket

Resolves [DEX-359](https://linear.app/rudderstack/issue/DEX-359/varsubst-package-fileresolver)

---

## Summary

Implements the FileResolver, which reads a single flat key-value YAML variable file and resolves variables from its contents. Each `--var-file` CLI argument (wired in a later ticket) will create a separate FileResolver instance added to the Substitutor's resolver chain.

---

## Changes

- Add `FileResolver` in the `resolver` package — reads flat YAML, parses into `map[string]string`, resolves via map lookup
- Non-string YAML scalars (int, bool, float, null) converted to string representations
- Nested structures (maps/arrays) rejected with `ErrVarFileParseFailed`
- Update `NewEnvResolver` signature to `(Resolver, error)` for consistency across all resolver constructors

---

## Testing

- 15 table-driven unit tests covering: missing file, invalid YAML, nested structures, type conversions, comments/empty lines, empty file, found/not-found lookups
- All existing EnvResolver tests updated and passing

---

## Risk / Impact

Low — new package with no callers yet; EnvResolver signature change is internal to the varsubst package which has no external consumers.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/varsubst/resolver/file_resolver.go
+++ b/cli/internal/varsubst/resolver/file_resolver.go
@@ -1,0 +1,53 @@
+package resolver
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/rudderlabs/rudder-iac/cli/internal/varsubst"
+	"gopkg.in/yaml.v3"
+)
+
+type fileResolver struct {
+	vars map[string]string
+}
+
+func NewFileResolver(path string) (Resolver, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("%w: %s", varsubst.ErrVarFileNotFound, path)
+		}
+		return nil, fmt.Errorf("reading variable file %s: %w", path, err)
+	}
+
+	var raw map[string]any
+	if err := yaml.Unmarshal(data, &raw); err != nil {
+		return nil, fmt.Errorf("%w: %s", varsubst.ErrVarFileParseFailed, path)
+	}
+
+	vars := make(map[string]string, len(raw))
+	for key, val := range raw {
+		switch v := val.(type) {
+		case string:
+			vars[key] = v
+		case int:
+			vars[key] = fmt.Sprint(v)
+		case float64:
+			vars[key] = fmt.Sprint(v)
+		case bool:
+			vars[key] = fmt.Sprint(v)
+		case nil:
+			vars[key] = ""
+		default:
+			return nil, fmt.Errorf("%w: key %q has non-scalar value in %s", varsubst.ErrVarFileParseFailed, key, path)
+		}
+	}
+
+	return &fileResolver{vars: vars}, nil
+}
+
+func (r *fileResolver) Resolve(name string) (string, bool) {
+	value, found := r.vars[name]
+	return value, found
+}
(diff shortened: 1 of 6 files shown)


cli/internal/varsubst/resolver/file_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+
+type fileResolver struct {
+	vars map[string]string
+}
+
+func NewFileResolver(path string) (Resolver, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("%w: %s", varsubst.ErrVarFileNotFound, path)
```
abhimanyubabbar: Why not absorb the Sentinel Errors here in this package i.e. resolver ?

This might create dependency issues like deadlocks in import.

cli/internal/varsubst/resolver/file_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+		case int:
+			vars[key] = fmt.Sprint(v)
+		case float64:
+			vars[key] = fmt.Sprint(v)
+		case bool:
+			vars[key] = fmt.Sprint(v)
+		case nil:
+			vars[key] = ""
+		default:
+			return nil, fmt.Errorf("%w: key %q has non-scalar value in %s", varsubst.ErrVarFileParseFailed, key, path)
```
abhimanyubabbar: Why should we silently treat nil as empty strings ?
Why not simply error out ?

cli/internal/varsubst/resolver/file_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+	for key, val := range raw {
+		switch v := val.(type) {
+		case string:
+			vars[key] = v
+		case int:
+			vars[key] = fmt.Sprint(v)
+		case float64:
+			vars[key] = fmt.Sprint(v)
+		case bool:
+			vars[key] = fmt.Sprint(v)
```
abhimanyubabbar, line 45: Maybe we need to lock it down in future that how should we treat the case where we have env keys with type conflicted values ?

Like 

db_host: "abc.postgres.com"
db_host: 5432 # maybe by mistake this was added

cli/internal/varsubst/resolver/file_resolver.go
abhimanyubabbar: I would name the errors a bit generic like ErrNotFound, ErrInternalError, ErrIllegalArgument exception.

The error then with proper context can make sure we can use these error types widely.

dhawal1248: will do

cli/internal/varsubst/resolver/file_resolver.go
dhawal1248: will do: do not allow nil values.nil values would only be hit in two cases - value of the env is not defined or id the value of the env is null.We wont support both cases as of now, document this via a comment.To set an env to empty value,use double/single quotes ""/''

cli/internal/varsubst/resolver/file_resolver.go
dhawal1248: will do: optional

cli/internal/varsubst/resolver/file_resolver.go
dhawal1248: discussed and resolved

abhimanyubabbar (approved):

PR #591, opened 2026-05-27
feat: add Substitutor core engine for variable substitution

dhawal1248:
## 🔗 Ticket

Resolves [DEX-360](https://linear.app/rudderstack/issue/DEX-360/varsubst-package-substitutor-core-engine)

---

## Summary

Implements the Substitutor — the core byte-scanning engine that finds `{{ .VAR }}` tokens in raw YAML bytes, resolves them through an ordered chain of resolvers, and returns substituted bytes. Includes comment detection, empty value handling, variable name validation, position computation, and diagnostic conversion.

---

## Changes

- Add `Substitutor` interface and `NewSubstitutor` constructor accepting variadic `Resolver` arguments
- Implement `SubstituteBytes` with reverse-iteration to preserve byte offsets during replacement
- Add `isInComment` for quote-aware YAML comment detection
- Add `parseVarName` to validate dot prefix and `[A-Za-z_][A-Za-z0-9_]*` name pattern
- Add single-pass `computePositions` to convert byte offsets to line/col from original bytes
- Add `ToDiagnostics` to bridge `[]SubstitutionError` into `validation.Diagnostics`
- Broad `varRegex` matches any `{{ content }}` token; validation happens in code

---

## Testing

- 34 unit tests covering substitution, comment detection, error positions, and diagnostics
- Covers: single/multiple vars, defaults, resolver priority, comment skipping, empty value quoting, whitespace variants, undefined vars, invalid names, missing dot prefix, mixed errors

---

## Risk / Impact

Low — new additive code, no changes to existing functionality.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/varsubst/substitutor.go
+++ b/cli/internal/varsubst/substitutor.go
@@ -1,0 +1,243 @@
+package varsubst
+
+import (
+	"fmt"
+	"regexp"
+	"slices"
+
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation"
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation/pathindex"
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation/rules"
+)
+
+type Resolver interface {
+	Resolve(name string) (value string, found bool)
+}
+
+// Matches any {{ content }} token. Group 1 captures the token (anything that
+// isn't whitespace, pipe, or closing brace). Group 2 (optional) captures the
+// default value after pipe — cannot contain "}" and surrounding whitespace is
+// stripped by the enclosing \s* groups. Validation of the token (dot prefix,
+// variable name pattern) happens in code after matching.
+var varRegex = regexp.MustCompile(`\{\{\s*([^}\s|]+)(?:\s*\|\s*([^}]*?))?\s*\}\}`)
+
+var validVarName = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
+
+type Substitutor interface {
+	SubstituteBytes(data []byte) ([]byte, []SubstitutionError)
+}
+
+type substitutor struct {
+	resolvers []Resolver
+}
+
+func NewSubstitutor(resolvers ...Resolver) Substitutor {
+	return &substitutor{resolvers: resolvers}
+}
+
+// rawError stores the name of the variable, the byte offset, and the error kind.
+// This is used to compute the line/column positions of the error.
+type rawError struct {
+	name   string
+	offset int
+	err    error
+}
+
+// SubstituteBytes finds all {{ .VAR }} tokens and resolves them through the
+// resolver chain. Matches are processed last-to-first so that replacing a token
+// with a different-length value does not invalidate earlier byte offsets.
+// A copy of the original bytes is kept so that error positions can be computed
+// against unmodified content after all replacements are done.
+func (s *substitutor) SubstituteBytes(data []byte) ([]byte, []SubstitutionError) {
+	matches := varRegex.FindAllSubmatchIndex(data, -1)
+	if len(matches) == 0 {
+		return data, nil
+	}
+
+	original := make([]byte, len(data))
+	copy(original, data)
+
+	var rawErrors []rawError
+
+	for i := len(matches) - 1; i >= 0; i-- {
+		match := matches[i]
+		matchStart, matchEnd := match[0], match[1]
+
+		if isInComment(data, matchStart) {
+			continue
+		}
+
+		token := string(data[match[2]:match[3]])
+
+		varName, err := parseVarName(token)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


cli/internal/varsubst/substitutor.go
```diff
@@ -0,0 +1,243 @@
...
+	"fmt"
+	"regexp"
+	"slices"
+
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation"
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation/pathindex"
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation/rules"
+)
+
+type Resolver interface {
```
abhimanyubabbar, line 14: We define the Resolver interface here as well ?

cli/internal/varsubst/substitutor.go
```diff
@@ -0,0 +1,243 @@
...
+	offset int
+	err    error
+}
+
+// SubstituteBytes finds all {{ .VAR }} tokens and resolves them through the
+// resolver chain. Matches are processed last-to-first so that replacing a token
+// with a different-length value does not invalidate earlier byte offsets.
+// A copy of the original bytes is kept so that error positions can be computed
+// against unmodified content after all replacements are done.
+func (s *substitutor) SubstituteBytes(data []byte) ([]byte, []SubstitutionError) {
```
abhimanyubabbar, line 66: Would we be able to differentiate replacing

```
{{name}}: "name"
```

As I think above one would be a valid YAML as well.

cli/internal/varsubst/substitutor.go
```diff
@@ -0,0 +1,243 @@
...
+type Resolver interface {
+	Resolve(name string) (value string, found bool)
+}
+
+// Matches any {{ content }} token. Group 1 captures the token (anything that
+// isn't whitespace, pipe, or closing brace). Group 2 (optional) captures the
+// default value after pipe — cannot contain "}" and surrounding whitespace is
+// stripped by the enclosing \s* groups. Validation of the token (dot prefix,
+// variable name pattern) happens in code after matching.
+var varRegex = regexp.MustCompile(`\{\{\s*([^}\s|]+)(?:\s*\|\s*([^}]*?))?\s*\}\}`)
```
abhimanyubabbar: Let's use the below mechanism to write defaults

{{ .VAR | default "abc" }} 

This will be in-line with go templating for future needs as well

abhimanyubabbar: As the current mechanism of simply using a pipe as a delimiter is ambiguous and confusing. We can borrow from  Sprig / Helm of using "default" in it.

abhimanyubabbar (commented):
Claude Feedback on the SubstituteBytes Function

 ---
  Issues That Cause Wrong Substitution

  1. Empty-string auto-quoting in pre-quoted contexts

  // line 106-108
  if resolved == "" {
      resolved = `""`
  }

  # Input
  value: "{{ .VAR }}"
  # VAR resolves to "" → resolved becomes ""

  # Output: value: """"   ← invalid YAML

  Same with single quotes: '{{ .VAR }}' → '""' — inserts literal "" inside single-quoted string.

  2. Substitution inside YAML quoted strings where {{ is literal text

  description: "Deploy uses {{ .Branch }} syntax"

  Regex matches {{ .Branch }}, resolves or errors. But user meant literal text inside a YAML string — no substitution should happen. The function has no
  YAML quoting awareness beyond comment detection.

  3. isInComment fails on escaped quotes in YAML double-quoted strings

  key: "val\"ue" # {{ .VAR }}

  Quote tracker sees \" as close-quote, then " before # as re-open. Thinks # is inside quotes → returns false → substitutes a token that's inside a
  comment.

  Reverse case also possible — token outside comment treated as inside comment due to mistracked quotes.

  4. Default values containing } produce wrong match boundaries

  pattern: {{ .RE | [a-z]{3} }}

  [^}]*? in regex stops at } in {3}. Match boundary shifts — either wrong default captured or regex fails to match the full {{ }} token, leaving it
  unsubstituted or partially matched.

  5. Resolved values with newlines corrupt YAML structure

  name: {{ .DESC }}
  port: 8080

  If DESC resolves to "line one\nline two":

  name: line one
  line two      ← now a top-level YAML key or parse error
  port: 8080

  Silently breaks all YAML below the injection point.

  6. Resolved values with YAML-special content change types

  enabled: {{ .FLAG }}
  count: {{ .NUM }}
  value: {{ .DATA }}

  If FLAG resolves to true → YAML boolean, not string.
  If NUM resolves to 123 → YAML integer, not string.
  If DATA resolves to null → YAML null, not string.
  If DATA resolves to [a, b] → YAML sequence, not string.

  No quoting/escaping applied — resolved value's YAML interpretation is uncontrolled.

dhawal1248:
will do: check the comment and evaluate if need any fixes - https://github.com/rudderlabs/rudder-iac/pull/591#pullrequestreview-4372314943

abhimanyubabbar (approved):

PR #592, opened 2026-05-27
feat(project): wire varsubst.Substitutor into project loading (DEX-361)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-361](https://linear.app/rudderstack/issue/DEX-361/project-integration-withsubstitutor-option-and-parsespecs-changes)

---

## Summary

Connects the `varsubst.Substitutor` engine (DEX-358/359/360) to the project loading pipeline. When a substitutor is configured, raw spec bytes pass through `SubstituteBytes` before `yaml.Decode`; any resolution errors surface as `validation.Diagnostics` and the offending spec is skipped. When no substitutor is provided (the existing default), behavior is identical to today.

Stacked PR — targets `feature/dex-360-...` so the diff stays scoped to DEX-361. Retarget to `main` once DEX-360 merges.

---

## Changes

- Add `project.WithSubstitutor(varsubst.Substitutor) ProjectOption` and a `substitutor` field on the project struct.
- In `parseSpecs`, run substitution before `RawSpec.Parse()`; on errors, convert via `varsubst.ToDiagnostics` and skip the spec; on success, build a fresh `RawSpec` to avoid leaving stale `parsed`/`pathIndex` cache on the original.
- Change `app.Deps.NewProject` to `NewProject(opts ...project.ProjectOption)` — backwards-compatible variadic; existing zero-arg callers unaffected.

---

## Testing

- Unit / integration tests at the project layer (`TestProject_Load_WithSubstitutor`) covering: variable resolution, non-string scalar type preservation (`port: 5432` parsed as int), undefined-variable diagnostic and spec exclusion, nil-substitutor unchanged behavior, mixed batch where the clean spec still loads despite an errored sibling.
- `make lint` clean, `make test` passes.

---

## Risk / Impact

Low. Feature is opt-in via `WithSubstitutor`; nil substitutor keeps existing call paths byte-identical.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/project/project.go
+++ b/cli/internal/project/project.go
@@ -225,7 +225,7 @@ func (p *project) parseSpecs(raw map[string]*specs.RawSpec) (map[string]*specs.R
 		if p.substitutor != nil {
 			substituted, subErrs := p.substitutor.SubstituteBytes(rawSpec.Data)
 			if len(subErrs) > 0 {
-				diags = append(diags, varsubst.ToDiagnostics(path, subErrs)...)
+				diags = append(diags, substitutionDiagnostics(path, subErrs)...)
 				continue
 			}
 			rawSpec = &specs.RawSpec{Data: substituted}
--- a/cli/internal/project/project.go
+++ b/cli/internal/project/project.go
@@ -15,6 +15,7 @@ import (
 	"github.com/rudderlabs/rudder-iac/cli/internal/validation/pathindex"
 	"github.com/rudderlabs/rudder-iac/cli/internal/validation/renderer"
 	"github.com/rudderlabs/rudder-iac/cli/internal/validation/rules"
+	"github.com/rudderlabs/rudder-iac/cli/internal/varsubst"
 )
 
 var log = logger.New("project")
@@ -45,6 +46,7 @@ type project struct {
 	specs            map[string]*specs.Spec
 	validationEngine validation.ValidationEngine
 	renderer         renderer.Renderer
+	substitutor      varsubst.Substitutor
 }
 
 // ProjectOption defines a functional option for configuring a Project.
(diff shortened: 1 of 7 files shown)


cli/internal/project/project.go
```diff
@@ -212,6 +222,15 @@ func (p *project) parseSpecs(raw map[string]*specs.RawSpec) (map[string]*specs.R
 	// only works on the data provided.
 	for path, rawSpec := range raw {
 
+		if p.substitutor != nil {
+			substituted, subErrs := p.substitutor.SubstituteBytes(rawSpec.Data)
+			if len(subErrs) > 0 {
+				diags = append(diags, varsubst.ToDiagnostics(path, subErrs)...)
```
dhawal1248: will do: create a new function that converts subErrors to diagnostic errors in a new file utils.go in the project package..do not do inside varsubst

cli/internal/project/project.go
```diff
@@ -212,6 +222,15 @@ func (p *project) parseSpecs(raw map[string]*specs.RawSpec) (map[string]*specs.R
...
+				continue
+			}
+			rawSpec = &specs.RawSpec{Data: substituted}
+		}
```
abhimanyubabbar: Let's just return from here because if the substitution never happened completely, it would result into unnecessary semantic validations being hit because maybe the referenced resource was never created. 

So it might be better to raise the errors till substitution stage and then the syntactic + semantic validation errors ?

abhimanyubabbar (changes requested):

cli/internal/project/project.go
dhawal1248: yeah makes sense, will do

abhimanyubabbar (approved):

PR #594, opened 2026-05-27
feat(cli): wire --var-file flag and experimental gate for variable substitution (DEX-362)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-362](https://linear.app/rudderstack/issue/DEX-362/cli-wiring-var-file-flag-experimental-gate-and-prerune-setup)

---

## Summary

Adds the user-facing entry point for variable substitution: a repeatable `--var-file` flag on `apply` and `validate`, plus a new `enableVarSubstitution` experimental flag that gates the whole mechanism. When the flag is off (default), behaviour is identical to today.

---

## Changes

- `config`: add `enableVarSubstitution` experimental flag (env `RUDDERSTACK_X_ENABLE_VAR_SUBSTITUTION`)
- `app`: add `NewProjectOptions(cfg, varFiles)` that assembles `[]project.ProjectOption` — currently wires the substitutor (`EnvResolver` first, then a `FileResolver` per `--var-file` in order) when the experimental flag is on; designed so future capabilities slot in alongside it
- `apply`, `validate`: register `--var-file` (`StringArrayVar`) and pass the assembled options to `deps.NewProject`
- **Deviation from ticket**: `migrate` deliberately does **not** get `--var-file`. Migration writes the loaded project back to disk, so substituting at load time would silently strip the user's template variables from the migrated YAML — almost certainly not the intended outcome for someone migrating a templated v0.1 project. Open to revisiting if migrate gains a "preserve templates" pathway later.
- `destroy` correctly remains untouched — no specs are loaded.

---

## Testing

- Unit tests on `NewProjectOptions`: flag-off short-circuit, flag-on with/without var files, `ErrVarFileNotFound`, `ErrVarFileParseFailed`
- Unit tests on `buildSubstitutor` resolver chain: env always wired, env-over-file priority, earlier-file-over-later-file priority, file resolver supplies values not in env
- `make lint` and `make test` clean

---

## Risk / Impact

Low — feature is gated behind an experimental flag that defaults to off, so the default code path is unchanged.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/app/options.go
+++ b/cli/internal/app/options.go
@@ -1,0 +1,49 @@
+package app
+
+import (
+	"fmt"
+
+	"github.com/rudderlabs/rudder-iac/cli/internal/config"
+	"github.com/rudderlabs/rudder-iac/cli/internal/project"
+	"github.com/rudderlabs/rudder-iac/cli/internal/varsubst"
+	"github.com/rudderlabs/rudder-iac/cli/internal/varsubst/resolver"
+)
+
+// NewProjectOptions assembles the project options that should be applied to
+// every project created via Deps.NewProject. Each capability is gated by its
+// own experimental flag in cfg; when a flag is off, the related option is
+// omitted entirely. Additional capabilities should be wired in here so all
+// command call sites pick them up uniformly.
+func NewProjectOptions(cfg config.Config, varFiles []string) ([]project.ProjectOption, error) {
+	var opts []project.ProjectOption
+
+	if cfg.ExperimentalFlags.EnableVarSubstitution {
+		sub, err := buildSubstitutor(varFiles)
+		if err != nil {
+			return nil, err
+		}
+		opts = append(opts, project.WithSubstitutor(sub))
+	}
+
+	return opts, nil
+}
+
+// buildSubstitutor wires the standard resolver chain: env resolver first
+// (highest priority), then a FileResolver per varFile in the order provided.
+func buildSubstitutor(varFiles []string) (varsubst.Substitutor, error) {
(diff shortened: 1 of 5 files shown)


cli/internal/app/options.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, fmt.Errorf("initialising env resolver: %w", err)
+	}
+
+	resolvers := []varsubst.Resolver{envR}
+	for _, path := range varFiles {
+		r, err := resolver.NewFileResolver(path)
+		if err != nil {
+			return nil, fmt.Errorf("initialising file resolver: %w", err)
+		}
+		resolvers = append(resolvers, r)
```
abhimanyubabbar, line 45: As per this logic, the file resolvers will be added in the order they are provided. Should the order be reversed because overrides file is provided afterwards and the variable if present in the overrides would be conidered with higher precedence than in base file ?

cli/internal/cmd/project/validate/validate.go
```diff
@@ -44,7 +46,12 @@ func NewCmdValidate() *cobra.Command {
 				return fmt.Errorf("initialising dependencies: %w", err)
 			}
 
-			p = deps.NewProject()
+			projectOpts, err := app.NewProjectOptions(config.GetConfig(), varFiles)
```
abhimanyubabbar, line 49: Do we reverse the order of var-files here ?

abhimanyubabbar (approved):
Minor comments mainly on the order of var files,

PR #601, opened 2026-05-29
doc(varsubst): add user-facing README for variable substitution (DEX-385)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-385](https://linear.app/rudderstack/issue/DEX-385/varsubst-add-user-facing-readme-documentation)

---

## Summary

Adds `cli/internal/varsubst/README.md` — a single, authoritative reference for the variable substitution feature, co-located with the implementation so it's discoverable from the package. The PRD/LLD have drifted from the actual implementation in a few load-bearing places (token syntax, default operator, var-file precedence, gate name), so every claim was verified against the source rather than the design docs.

---

## Changes

- New `cli/internal/varsubst/README.md` covering:
  - Quick start (enable → placeholder → var file → apply)
  - Enabling the feature via env vars / config file / `experimental enable` command — calls out that both `experimental: true` AND the flag must be on
  - Syntax: `{{ .VAR }}` with leading dot, `{{ .VAR | default }}` pipe form
  - Value resolution priority table (env > later var-file > earlier var-file > inline default)
  - Supported commands (apply + validate only — not destroy/migrate/import)
  - Types and quoting guidance
  - Error handling with a real example of the renderer's `-->`/caret output
  - Security best practices
  - FAQ / edge cases (no escape syntax, non-recursive, `#` in quoted strings, type preservation, empty values, embedded code blocks, file path resolution)
  - Callout that the `RUDDER_` prefix applies only to env vars, NOT to var-file keys

---

## Testing

Documentation-only; no code paths touched.

- Verified every load-bearing claim against the source: `substitutor.go`, `resolver/`, `app/options.go`, `cmd/apply/apply.go`, `cmd/validate/validate.go`.
- Confirmed the example error output matches the actual renderer format (DEX-360).
- Confirmed the documented var-file precedence matches DEX-384's reversal (later wins).

---

## Risk / Impact

Low — docs only, no runtime impact.

Notes: README flags two pieces of follow-up cleanup that are intentionally **out of scope** here:
1. The \`--var-file\` flag help text in \`apply.go\` / \`validate.go\` still reads \"earlier files take priority\" — stale after DEX-384. Small separate fix.
2. Eventual move to a top-level user-facing docs location can be revisited later.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated (N/A — docs only)


cli/internal/varsubst/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+```yaml
+host: "{{ .DB_HOST }}"
+```
+
+- The token is `{{ .NAME }}` — the **leading dot is required** (`{{ DB_HOST }}` is invalid).
+- Whitespace inside the braces is flexible: `{{.DB_HOST}}`, `{{ .DB_HOST }}`, and
+  `{{  .DB_HOST  }}` are all equivalent.
+- Variable names must match `[A-Za-z_][A-Za-z0-9_]*` — start with a letter or underscore,
+  followed by letters, digits, or underscores. (So `{{ .123 }}` or `{{ .my-var }}` are
```
sivashanmukh: What is wrong with `{{ .my-var }}`

dhawal1248: The hyphen is the problem. Variable names follow identifier rules — `[A-Za-z_][A-Za-z0-9_]*` (letters, digits, underscores; must start with a letter or underscore), the same rule as environment variable names. `-` isn't allowed, so `{{ .my-var }}` fails with `invalid variable syntax` (see `substitutor.go:21` + `parseVarName`). It's intentional and consistent with env-var naming — you can't export `RUDDER_my-var` either. Use `{{ .my_var }}` instead.

I've reworded the README to spell this out with valid/invalid examples — done in bad94855.

sivashanmukh (approved):
Lgtm

PR #603, opened 2026-06-01
feat: import manifest provider and classifier

Jayachand:
## 🔗 Ticket

Resolves [DEX-388](https://linear.app/rudderstack/issue/DEX-388/add-spec-classifier-and-import-manifest-provider-skeleton)

---

## Summary

PR 2 of the import-manifest workflow (see `docs/superpowers/specs/2026-06-14-import-manifest-pr-plan.md`). It adds the project-level import-manifest provider + spec classifier **and wires them into the project load dispatch** so an `import-manifest` spec is classified as project-level, routed to the dedicated provider (outside the resource/CompositeProvider tree), and its workspace entries are accumulated. Behaviour for resource specs is unchanged.

Note: the manifest provider's match patterns/rules are intentionally **not** registered in the validation registry yet — that wiring (so `import-manifest` is a known kind to `SpecSyntaxValidRule`) lands in the next PR, alongside the manifest validation rules.

---

## Changes

- `cli/internal/project/specs/classifier.go` — `KindImportManifest` constant, `SpecLevel` enum, `SpecClassifier.Classify` (resource-level vs project-level; internal, not in the YAML schema).
- `cli/internal/project/importmanifest/manifest.go` — `parseWorkspaces`, strict-decodes `spec.workspaces` and rejects unknown fields at every level.
- `cli/internal/project/importmanifest/provider.go` — project-level `Provider`: `LoadSpec` appends workspace entries, `ImportManifest()` returns a copy (nil when empty), declares the import-manifest kind/match-pattern, rule sets nil until a later PR, and `RuleDocEntries()` (so it satisfies `ProjectProvider`).
- `cli/internal/project/project.go` — **NEW wiring**: `project` holds an `importManifestProvider` + `SpecClassifier` (created in `New`); `loadSpec` classifies each spec and routes project-level specs to the manifest provider, leaving the resource version-dispatch unchanged.

---

## Testing

- Unit tests:
  - classifier — `Classify` over manifest / resource / empty / legacy kinds
  - manifest — `parseWorkspaces` happy path, empty, unknown-field rejection (top + nested)
  - provider — type surfaces, append behaviour, malformed rejection, nil-when-empty, copy-safety, stubs
  - **project (new)** — `Load` routes an `import-manifest` spec to the manifest provider (not the resource provider) and accumulates its workspace entry
- `go build ./cli/internal/project/...` clean; targeted project tests pass.

---

## Risk / Impact

Low — resource-spec handling is unchanged; the new dispatch arm only diverts the `import-manifest` kind, which no production import emits yet.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/project/specs/classifier.go
```diff
@@ -0,0 +1,26 @@
+package specs
+
+// KindImportManifest is a project-level kind handled outside the resource
+// provider tree, so the constant lives with the classifier rather than
+// alongside resource kinds.
+const KindImportManifest = "import-manifest"
```
abhimanyubabbar: Shouldn't importmanifest be source of truth for the kind which will be representing it ?

cli/internal/project/project.go
```diff
@@ -126,6 +129,13 @@ func (p *project) Specs() map[string]*specs.Spec {
 }
 
 func (p *project) loadSpec(path string, spec *specs.Spec) error {
+	// Project-level specs (e.g. import-manifest) are handled outside the resource
+	// provider tree by their dedicated provider; resource-level specs flow through
+	// the version-based dispatch below.
+	if specs.Classify(spec) == specs.ProjectSpec {
+		return p.importManifestProvider.LoadSpec(path, spec)
+	}
```
abhimanyubabbar: For now, it's okay given import manifest provider is the only ProjectProvider but similar to the ResourceProvider ( Composite Structure ) eventually we might end up with Project Level Spec Provider which is composite


```suggestion
	if specs.ProjectSpec(spec) {
	    // projectProvider is an amalgamation of
	    // all the project level spec providers
	    // 
		return p.projectProvider.LoadSpec(path, spec)
	}
```

cli/internal/project/project.go
```diff
@@ -40,14 +41,15 @@ type Project interface {
...
-	provider         ProjectProvider
-	loader           Loader
-	workspaceID      string
-	specs            map[string]*specs.Spec
-	validationEngine validation.ValidationEngine
-	renderer         renderer.Renderer
-	substitutor      varsubst.Substitutor
+	location               string
+	provider               ProjectProvider
+	importManifestProvider *importmanifest.Provider
```
abhimanyubabbar: Should we use it as ProjectProvider interface only ?

cli/internal/project/importmanifest/provider.go
```diff
@@ -0,0 +1,89 @@
...
+// RuleDocEntries returns nil — the manifest provider contributes no doc fragments
+// until its validation rules land. Present so the provider satisfies RuleProvider
+// (and therefore ProjectProvider).
+func (p *Provider) RuleDocEntries() []docs.RuleDocEntry {
+	return nil
+}
+
+// ImportManifest returns a copy of the aggregated manifest, or nil when no
+// manifest was loaded. The copy keeps callers from corrupting append-only state.
+func (p *Provider) ImportManifest() *specs.WorkspacesImportMetadata {
```
abhimanyubabbar: Where is it getting used ?

cli/internal/project/importmanifest/provider.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (p *Provider) SupportedTypes() []string {
+	return nil
+}
+
+func (p *Provider) SupportedMatchPatterns() []rules.MatchPattern {
+	return []rules.MatchPattern{
+		rules.MatchKindVersion(specs.KindImportManifest, specs.SpecVersionV1),
+	}
+}
```
abhimanyubabbar, line 41: We are not adding them in BuildRegistry function right ? Because supported match patterns are used only from the custom provider as of now right ?

cli/internal/project/importmanifest/provider.go
abhimanyubabbar: Where are we using this function ?

abhimanyubabbar (approved):
Overall the PR looks good with some queries and suggestions

cli/internal/project/specs/classifier.go
Jayachand: moved to manifest

cli/internal/project/project.go
Jayachand: agree. will update interface when accessor method is used in code

cli/internal/project/importmanifest/provider.go
Jayachand: It is used in handleValidation method. This method returns list of manifest workspaces to be broadcasted to composite provider. Removing this until its actual implementation

cli/internal/project/importmanifest/provider.go
Jayachand: It will be added in build registry in upcoming prs

cli/internal/project/importmanifest/provider.go
Jayachand: in handlevalidation function

PR #471, opened 2026-03-19
feat: added foundation for rule docs

abhimanyubabbar:
## 🔗 Ticket

Resolves [DEX-269](https://linear.app/rudderstack/issue/DEX-269/phase-1-docs-package-foundation)

---

## Summary

Creates `cli/internal/validation/docs` — a fully self-contained package with zero changes to existing code. This is Phase 1 of the DEX-216 RuleDoc Generator epic. It establishes the foundational types, YAML fragment loader, and catalog validator that all later phases (DEX-270–DEX-273) build on.

---

## Changes

- `types.go` — all authored (`RuleDocEntry`, `MatchBehaviorEntry`, `MatchPatternDoc`, `ValidExample`, `InvalidExample`, `ExpectedDiagnostic`) and resolved (`RulesDoc`, `ToolMetadata`, `ResolvedRule`) types with `validate` tags
- `rules_doc_entry.go` — `LoadFragment(fsys fs.FS, dir string)` reads all `.yaml`/`.yml` files from a directory, unmarshals each into `RuleDocEntry`, skips non-YAML files and subdirectories
- `rules_doc.go` — `RulesDoc.Validate(registeredRuleIDs []string)` enforces 4 rules: structural field validation via `go-playground/validator`, `applies_to` coverage check, unique `example_id` per rule, and registered completeness (no missing or orphan entries)
- `rules_doc_entry_test.go` — unit tests for `LoadFragment` covering 7 behaviours (nonexistent dir, empty dir, single/multiple YAML, skip non-YAML, skip subdirs, malformed YAML)
- `rules_doc_test.go` — table-driven unit tests for `RulesDoc.Validate` with one sub-table per validation rule, covering all valid and invalid scenarios

---

## Testing

- Unit tests added for all new logic — `go test ./cli/internal/validation/docs/...`
- 97.3% statement coverage
- `make lint` — 0 issues
- `make test` — full suite green, no existing tests affected

---

## Risk / Impact

Low — no existing files modified, purely additive new package.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated
- [x] No breaking changes (or documented)

---

> [!NOTE]
> **Low Risk**
> Purely additive new `cli/internal/validation/docs` package with unit tests; no existing code paths are modified, so risk is limited to future consumers relying on these new validation rules.
> 
> **Overview**
[...]

PR #604, opened 2026-06-02
feat: dex 271 phase 3 gen rule docs script skeleton

sivashanmukh:
## 🔗 Ticket

Resolves [DEX-271](https://linear.app/rudderstack/issue/DEX-271) — Phase 3: generator + serializer + CLI command.
Sub-issues: DEX-396, DEX-393, DEX-394, DEX-395, DEX-401.

> Stacked on #471 (docs foundation, DEX-269) and #475 (registry + RuleProvider, DEX-270) — merge those first; this branch is cut from the #475 branch.

---

## Summary

Phase 3 of the rule-docs pipeline: the engine that turns a provider's live validation rules plus authored doc fragments into a canonical, validated catalog, and the `rudder-cli docs rules` command that renders it. Authoring of the fragments themselves lands in Phase 4 (#607).

This branch was rewritten into a clean, incremental history — each commit builds and tests independently, with no add-then-remove churn.

---

## Changes

Four commits, each independently testable:

1. **`feat(validation/docs): add RuleDoc generator`** — resolve provider rule metadata + authored fragments into a `DocumentedRules` catalog, validating struct / applies-to coverage / unique-example / registered-completeness invariants.
2. **`feat(validation/docs): add RulesDoc serializer (yaml + json)`** — render the catalog to canonical YAML and JSON.
3. **`feat(cmd/docs): add docs rules command`** — extract `BuildRegistry` so project loading and the docs command observe an identical rule set; render via `--format`.
4. **`chore: add make gen-rule-docs target`** — regenerate the catalog from authored fragments.

> Note: an earlier revision added an opt-in `--layout by-version` grouped renderer (`grouped.go`) and then removed it in a later commit. That do/undo churn (~353 lines) has been dropped from the history entirely — the feature was never in the final state.

---

## Testing

- Unit tests for the generator, serializer, and command (`make test`).
- `make lint` clean.

---

## Risk / Impact

Low. New, self-contained docs subsystem + a read-only CLI command; no change to the apply cycle.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


Makefile
```diff
@@ -33,6 +33,12 @@ build:
 clean:
 	rm -rf bin
 
+RULE_DOCS_OUTPUT_DIR ?= docs/generated
+
+.PHONY: gen-rule-docs
+gen-rule-docs: ## Generate the validation rule documentation artifact
+	$(GO) run ./cli/cmd/rudder-cli docs rules --output-dir $(RULE_DOCS_OUTPUT_DIR)
```
abhimanyubabbar: As discussed can we move it to an internal script rather than a command on the tool ?

cli/internal/validation/docs/generator.go
```diff
@@ -0,0 +1,86 @@
...
+		resolved = append(resolved, resolveRule(r, "semantic", entryByID))
+		registeredRuleIDs = append(registeredRuleIDs, r.ID())
+	}
+
+	sort.Slice(resolved, func(i, j int) bool {
+		return resolved[i].RuleID < resolved[j].RuleID
+	})
+
+	doc := DocumentedRules{
+		SchemaVersion: 1,
```
abhimanyubabbar: Let's put it as const at the top ?

cli/internal/validation/docs/generator.go
```diff
@@ -0,0 +1,86 @@
...
+	// MatchBehavior stays nil when no authored entry exists — that's expected
+	// during the skeleton phase and surfaces as a validation error downstream.
+	var matchBehavior []MatchBehaviorEntry
+	if entry, ok := entryByID[ruleID]; ok {
+		matchBehavior = entry.MatchBehavior
+	}
+
+	return DocumentedRule{
+		RuleID:        ruleID,
+		Provider:      providerFromRuleID(ruleID),
```
abhimanyubabbar: I am just thinking if we even need the provider and what benefit would it give to expose this information in hugo docs to the caller ?

abhimanyubabbar: Should we remove the Provider field ?

cli/internal/validation/docs/serializer.go
```diff
@@ -0,0 +1,86 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+// Output formats for the rules catalog. YAML is consumed by Hugo / humans,
+// JSON by LLMs / tooling. Both carry identical snake_case keys.
+const (
+	FormatYAML = "yaml"
+	FormatJSON = "json"
+	FormatBoth = "both"
+)
```
abhimanyubabbar: Let's just focus on only yaml to be the output for the cli ?

cli/internal/validation/docs/serializer.go
```diff
@@ -0,0 +1,86 @@
...
+	if format == FormatJSON || format == FormatBoth {
+		// JSON is intentionally derived from the YAML bytes (not marshaled from
+		// the struct directly) — see yamlToJSON for the rationale.
+		jsonBytes, err := yamlToJSON(yamlBytes)
+		if err != nil {
+			return err
+		}
+		if err := writeFile(filepath.Join(outDir, "rules.json"), jsonBytes); err != nil {
+			return err
+		}
```
abhimanyubabbar: Maybe this function is simply part of the script that we are writing which gets the documented rules and based on the -o flag outputs them in yaml format.

The -o flag value could be controlled via the env for CI / CD.

In addition to this, it could be a simple go script in `/cmd/gen-rule-docs/script.go` and we can have it parse the inputs through the flags

abhimanyubabbar (changes requested):

cli/internal/validation/docs/generator.go
sivashanmukh: Good question. Provider isn't load-bearing for generation. It's derived purely from the rule-ID prefix (providerFromRuleID), so in one sense it's redundant data. The reason it's in the artifact is grouping: the Hugo docs want rules bucketed by product area, and this hands consumers a ready-made grouping key without parsing rule IDs.

Two concessions though:

"provider" is the wrong name to expose. It leaks an internal CLI concept, and for the gatekeeper rules (metadata-syntax-valid, duplicate-urn) it resolves to project, which isn't a provider at all. If we keep the field it should be reframed as a consumer-facing axis  {domain | sub-product | feature }. 

Grouping/presentation is meant to be the translation layer's job, and that layer can derive the same bucket from rule_id just as easily.

The options I see:
(a) drop it from canonical and let the translation layer derive the grouping, or 
(b) keep it but rename to a stable domain/feature axis. 

I lean to (b) because a named grouping key in the canonical contract is cheap and saves every consumer re-deriving it and more importantly qualifies it in our product context. But I'm happy either way. Which do you prefer?

cli/internal/validation/docs/serializer.go
sivashanmukh: Simplifies things. Happy to remove code :)

cli/internal/validation/docs/serializer.go
sivashanmukh: Agreed. I remember that conversation with @fxenik as well.

Let's move it off the CLI into a standalone cmd/gen-rule-docs script, invoked via the make gen-rule-docs target and CI (output dir via flag/env). No reason for a release-only artifact step to sit on the public CLI surface.

One thing I'll carry over from the command so we don't lose it: the generator must enumerate the exact same rule set the validator uses (that was the original reason it went through app.Deps). So rather than the script building its own registry, I'll extract a shared, auth-free constructor for the composite provider + BuildRegistry that both the script and the normal app path call. 

As I was exploring this path, I also realized that needing a command for this would force an auth flow which is unnecessary for this. Generation would then make no network calls, because routing through app.NewDeps() today forces a token even in CI, which is unnecessary.

Net: script + make target, no public command, shared construction so the rule set can't drift, and no auth needed to generate. Will push the change.

cli/internal/validation/docs/generator.go
sivashanmukh: We should definitely remove Provider. But the question is if we should replace it with something else or leave it completely out.

Please refer to my previous response.

abhimanyubabbar: Let's leave it completely out for now. Better to add later than expose something which exposes internal structure

abhimanyubabbar: Answered in the below comment

cli/cmd/gen-rule-docs/main.go
```diff
@@ -0,0 +1,75 @@
...
+		cp.RuleDocEntries(),
+		app.GetVersion(),
+		time.Now().UTC().Format(time.RFC3339),
+	)
+	if len(verrs) > 0 {
+		for _, verr := range verrs {
+			fmt.Fprintln(os.Stderr, verr)
+		}
+		return fmt.Errorf("rule docs validation failed: %d error(s)", len(verrs))
+	}
```
abhimanyubabbar: deps package already has the composite provider with it. Should we use it ?

cli/internal/validation/docs/generator.go
sivashanmukh: Got it.

cli/cmd/gen-rule-docs/main.go
abhimanyubabbar: The way I would do is:

deps, err := NewDeps()
// err check

* get compositeProvider from deps
* create a registry using provider
* assign the registry to project through ProjectOptions, if empty in caller, then the registry will be created by default from the project ( like overriding through options )
* use the same pattern in gen-rule-docs

PR #608, opened 2026-06-05
feat(secret): add self-redacting secret.String value type (DEX-407)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-407](https://linear.app/rudderstack/issue/DEX-407/secretstring-core-self-redacting-value-type-and-package)

---

## Summary

Introduces `secret.String`, a self-redacting value type in a new `cli/internal/secret` package. It makes sensitivity a property of a field's *type*: every standard Go surface that prints, formats, logs, or serializes the value returns a redacted form automatically, and the real value escapes only through one explicit, greppable accessor — `Reveal()`. The type is inert until a provider adopts it, so this PR has zero blast radius.

This is step 1 of the staged rollout in the sensitive-fields design (framework only). Provider adoption, diff/render integration, and the import flow are deliberately out of scope and tracked separately (DEX-408/409/410).

---

## Changes

- New `secret` package with the `String` type — opaque, comparable, no exported fields; the zero value is a valid "no secret yet".
- Constructors `New(string)` (real value) and `NewRedacted()` (remote-supplied placeholder); accessors `Reveal()`, `IsZero()`, `IsRedacted()`.
- Redacting surfaces, all routed through a single `masked()` helper: `fmt.Stringer`, `fmt.Formatter` (every verb), `fmt.GoStringer`, `slog.LogValuer`, `json.Marshaler`, and a YAML marshaler.
- Spec-reading surfaces: `UnmarshalYAML` / `UnmarshalJSON` read a bare string (errors wrapped with action context).
- Masking style: `****` + last 4 characters (rune-aware) when the value is ≥ 8 characters, `***` for shorter values, `(remote-redacted)` for remote placeholders.

---

## Testing

- Unit tests (`go test ./cli/internal/secret/`): table-driven over every redaction surface (`%v %s %q %x %+v %#v`, slog, JSON, YAML), `Reveal()` returning the real value even after rendering, `NewRedacted()` never leaking, in-struct marshal/unmarshal (export/spec paths), equality semantics, and rune-aware masking of multi-byte values.
- `go vet` clean; `golangci-lint` reports 0 issues.
- No E2E needed — the type is inert (no apply-cycle wiring) until a provider adopts it, per the design's rollout step 1.

---

## Risk / Impact

Low — new standalone package, imported by nothing yet, so it cannot affect existing behavior.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


sivashanmukh (commented):
Reviewed with a security lens given this is the secret foundation. The architecture is sound — unexported field, single `masked()` chokepoint, one greppable `Reveal()`, every fmt/log/serialize surface covered; I confirmed the repo is yaml.v3-only (so `UnmarshalYAML` won't be silently skipped) and the base HTTP client doesn't log request bodies.

Two items worth weighing before this becomes the base others build on:
- 🔴 The mask reveals the last 4 chars for any value ≥ 8 runes — up to 50% of a short secret (inline on `hintThreshold`).
- 🟡 The `IsZero` doc points adopters at a send-decision that, for a redacted value, silently wipes the remote secret — the policy actually lives in the differ (DEX-408/411/412); suggest softening the doc.

The rest are a rename, doc precision, and a defense-in-depth `MarshalText`. Nothing blocks the inert package from merging.

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+	// hintThreshold is the minimum number of characters before masked() reveals
+	// the last four as a spot-check hint; below it the value is fully masked so the
+	// hint never exposes a meaningful fraction of a short secret.
+	hintThreshold = 8
+)
+
+// String is an opaque, comparable holder for a sensitive value. The zero value
+// is a valid "no secret yet". It has no exported fields, so the real value can
+// only be obtained through Reveal().
+type String struct {
```
sivashanmukh, line 30: **Rename `String` → `SensitiveString` (or `secret.Value`).** No existing `Secret`/`Sensitive` type in the repo, so no conflict. At a use site, `Token secret.String` reads too close to a plain `string` alias; for a type whose entire purpose is safety, an unmistakable name beats the `sql.NullString`-style symmetry. `SensitiveString` is self-describing even unqualified.

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	redactedPlaceholder = "(remote-redacted)"
+	shortMask           = "***"
+	// hintThreshold is the minimum number of characters before masked() reveals
+	// the last four as a spot-check hint; below it the value is fully masked so the
+	// hint never exposes a meaningful fraction of a short secret.
+	hintThreshold = 8
```
sivashanmukh, line 24: 📝 Doc accuracy, not a behavior change — make the comment match the code. At exactly 8 runes `masked()` returns `****1234` (half the secret; `secret_test.go:27` covers it), so "never exposes a meaningful fraction" is wrong at the boundary. Also "runes", not "characters":

```suggestion
	// hintThreshold is the minimum length, in runes, at which masked() reveals the
	// last four runes as a spot-check hint; shorter values are fully masked ("***").
	// At exactly this length the tail is half the value — the threshold is the
	// shortest secret for which exposing a 4-rune tail is an acceptable leak.
	hintThreshold = 8
```

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+
+// masked is the single redacted representation used by every formatting surface.
+func (s String) masked() string {
+	if s.redacted {
+		return redactedPlaceholder
+	}
+	// Count and slice by rune so multi-byte values are measured by character and
+	// the hint never splits a rune.
+	if r := []rune(s.v); len(r) >= hintThreshold {
+		return "****" + string(r[len(r)-4:])
```
sivashanmukh, line 67: Make the rule explicit here too: it masks **everything** and reveals only the **last 4 runes**, and only when `len >= hintThreshold`; anything shorter (including empty) is **fully** masked.

```suggestion
	// Reveal only the last 4 runes, and only when the value is long enough that
	// the hint stays a small minority (see hintThreshold). Slice by rune so a
	// multi-byte value is measured by character and the hint never splits a rune.
	if r := []rune(s.v); len(r) >= hintThreshold {
		return "****" + string(r[len(r)-4:])
```

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Reveal returns the real value. This is the only escape hatch and every call
+// site is greppable, so revelations can be audited.
+func (s String) Reveal() string { return s.v }
+
+// IsZero reports whether there is no secret to send: an empty value that is not a
+// remote-redacted placeholder. Providers use it to decide whether to include the
+// field in an outgoing request. New("") equals the zero value, so an empty secret
+// is treated the same as "unset".
+func (s String) IsZero() bool { return s.v == "" && !s.redacted }
```
sivashanmukh, line 54: The "include in an outgoing request" guidance is a footgun: for `NewRedacted()`, `IsZero()` is `false` but `Reveal()` is `""`, so an adopter following it literally on a full-object (PUT) update sends `""` and wipes the remote secret. The "redacted => converged, don't send" policy lives in the differ ([DEX-408](https://linear.app/rudderstack/issue/DEX-408)) + conventions ([DEX-411](https://linear.app/rudderstack/issue/DEX-411)), rotation in [DEX-412](https://linear.app/rudderstack/issue/DEX-412). Soften the doc so it doesn't imply `IsZero` alone gates the request:

```suggestion
// IsZero reports whether there is no secret to send: an empty value that is not a
// remote-redacted placeholder. New("") equals the zero value, so an empty secret is
// treated the same as "unset". It does NOT by itself gate an outgoing request: a
// redacted value is non-zero yet must never be sent (Reveal returns ""); that
// "redacted => don't send" rule lives in the differ (DEX-408).
func (s String) IsZero() bool { return s.v == "" && !s.redacted }
```

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+// conversion use it.
+func New(v string) String { return String{v: v} }
+
+// NewRedacted marks a value the remote API returned as a placeholder rather than
+// the real secret.
+func NewRedacted() String { return String{redacted: true} }
+
+// Reveal returns the real value. This is the only escape hatch and every call
+// site is greppable, so revelations can be audited.
+func (s String) Reveal() string { return s.v }
```
sivashanmukh, line 48: `Reveal()` returns `""` for a redacted placeholder, so a caller that builds an outgoing body from it silently wipes the remote secret. Worth a one-line warning here:

```suggestion
// Reveal returns the real value. This is the only escape hatch and every call site
// is greppable, so revelations can be audited. It returns "" for a redacted
// placeholder, so never build an outgoing request body from the result — a redacted
// value must be skipped, not sent (see IsZero and DEX-408).
func (s String) Reveal() string { return s.v }
```

cli/internal/secret/secret.go
```diff
@@ -0,0 +1,115 @@
...
+// LogValue implements slog.LogValuer so structured logs mask automatically.
+func (s String) LogValue() slog.Value { return slog.StringValue(s.masked()) }
+
+// MarshalJSON redacts. This is load-bearing for export, which json.Marshals a
+// spec before turning it into YAML. Outgoing API request bodies use plain string
+// fields populated via Reveal(), so the real value still reaches the wire.
+func (s String) MarshalJSON() ([]byte, error) { return json.Marshal(s.masked()) }
+
+// MarshalYAML redacts any direct YAML serialization of a String.
+func (s String) MarshalYAML() (any, error) { return s.masked(), nil }
```
sivashanmukh, line 93: Add `MarshalText` too: some encoders/loggers check `encoding.TextMarshaler` before falling back to reflection on the unexported `v` — the one surface not covered today.

```suggestion
// MarshalYAML redacts any direct YAML serialization of a String.
func (s String) MarshalYAML() (any, error) { return s.masked(), nil }

// MarshalText redacts for encoding.TextMarshaler consumers that reflect past Stringer.
func (s String) MarshalText() ([]byte, error) { return []byte(s.masked()), nil }
```

sivashanmukh (changes requested):

cli/internal/secret/secret.go
sivashanmukh: Heads-up across the stack: #614 (DEX-408) goes the other way on this. It renames `redacted`->`unknown` and makes an unknown secret always re-apply, rather than treating it as converged — so the "don't send" framing in my suggestion above no longer matches the final direction (the differ re-applies instead of skipping). Worth reconciling this `IsZero` doc wording with #614's policy before either merges.

cli/internal/secret/secret.go
dhawal1248: I'd lean toward keeping `String`. The package qualifier is always present at use sites — `accessToken secret.String` — and Go convention leans on the package name rather than restating it (`sql.NullString`, `bytes.Buffer`, `strings.Reader`). `secret.SensitiveString` stutters (`secret` + `Sensitive`). If we move off `String`, `secret.Value` is the better alternative, but I think `String` is both idiomatic and signals it's string-backed. Happy to revisit if you feel strongly.

cli/internal/secret/secret.go
dhawal1248: Agreed on both counts — at length 8 the tail is half the value, not a "meaningful fraction", and it should say "runes". Reworking the comment to state the boundary explicitly and use "runes".

cli/internal/secret/secret.go
dhawal1248: Will spell out the contract: mask everything, reveal only the last 4 runes, and only when `len(runes) >= hintThreshold` (anything shorter, including empty, is fully masked). #614 already switches the length check to rune-slicing so the hint never splits a multi-byte rune.

cli/internal/secret/secret.go
dhawal1248: Good catch, and thanks for the #614 cross-link. #614 settles the direction: an unknown secret always re-applies, so the differ re-applies rather than skipping and never builds an outgoing body from a placeholder — `IsZero` doesn't gate the request on its own. I'll align the `IsZero` doc with that policy so the two PRs read consistently before either merges.

cli/internal/secret/secret.go
dhawal1248: Agreed. Adding a note on `Reveal` that it returns `""` for an unknown/redacted placeholder and that callers must skip such a value rather than send it, with a TODO that we may harden it to error instead of silently returning `""`.

cli/internal/secret/secret.go
dhawal1248: Skipping this one for now — `String()`/`Format`/`GoString`/`LogValue`/`MarshalJSON`/`MarshalYAML` already cover the surfaces we serialize and log through, and I'd rather not add an interface we don't exercise. Easy to add later if we hit a `TextMarshaler`-only encoder.

dhawal1248:
Added review comment changes to the #614 PR ( commit 07877b5facb18cf7f27dd74d468ef597d11f5375)

cli/internal/secret/secret.go
sivashanmukh: Sounds good to me. secret.String should be indicative enough.

cli/internal/secret/secret.go
sivashanmukh: The contract is now stated where it matters — the reworked hintThreshold doc in 07877b5 (on [#614](https://github.com/rudderlabs/rudder-iac/pull/614)) spells out mask-everything / last-4-runes / fully-masked-below-threshold. Good enough for me; duplicating it on masked() itself is optional. Resolving.

cli/internal/secret/secret.go
sivashanmukh: Encoders that fall past TextMarshaler to reflection cannot read unexported fields, so the realistic worst case is {}/an error, not a leak. Every surface we actually serialize through is covered and tested. We can let it pass

sivashanmukh (approved):

PR #614, opened 2026-06-09
feat(syncer): make the apply-cycle diff secret-aware (DEX-408)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-408](https://linear.app/rudderstack/issue/DEX-408/make-the-apply-cycle-diff-secret-aware)

> ⚠️ **Stacked on #608 (DEX-407).** Base is the DEX-407 branch so the diff shows only DEX-408's work; rebase onto `main` once #608 merges.

---

## Summary

Makes the apply-cycle diff secret-aware.

**Deliberate deviation from the ticket:** the ticket asked to treat a masked/placeholder remote secret as *converged* (no diff). Because the backend never returns secret values, the remote side is *always* a placeholder, so converging would mean a changed secret could never be applied. We therefore made an **unknown** secret *always diff* (re-applied every run), surfaced in a separate "always re-applied" section and excluded from the import "not synced" check. In effect this folds the deferred DEX-412 escape-hatch into the default behaviour, and renames the "redacted" concept to "unknown".

---

## Changes

- **`secret` package:** rename redacted→unknown (`NewUnknown`, `IsUnknown`, `(unknown)` placeholder); `Diff` returns true whenever either side is unknown.
- **Differ:** secret-aware comparison; `PropertyDiff`/`ResourceDiff.SecretOnly` classification (computed once during the existing build pass) and `Diff.HasNonSecretDiff()`; `*secret.String` survives the `RawData` struct→map decode like `*PropertyRef`.
- **Reporter:** secret-only resources listed in a separate "Always re-applied (secret values can't be read back)" section, rendered via dedicated helpers (`renderSecret` / `renderSecretDiff` / `renderUpdatedResources`); the generic diff-rendering algorithm stays type-agnostic.
- **Importer:** "project not synced" guard uses `HasNonSecretDiff()` so always-re-applied secrets don't permanently block imports.
- **Fix:** `collectReferencesByReflection` skips unexported fields (was panicking on `secret.String` carried in a `RawData` struct).

---

## Testing

Unit tests, `make lint` (clean), and full `make test` (passing):

- **secret package:** every redacting surface + the flipped `Diff`.
- **differ:** secret comparison, `SecretOnly` propagation through nested maps, struct→map pointer survival, `HasNonSecretDiff`.
- **reporter:** separate section, annotation line, and masking (no real value ever printed).

No E2E added: no provider wires `secret.String` in production yet (adoption is DEX-409); this lands the framework + policy.

---

## Risk / Impact

**Low.** Inert until a provider adopts `secret.String` (DEX-409). The cross-cutting changes are the differ's `CompareData` signature (now also returns a secret-only bool — internal only, all callers updated) and the `references.go` reflection guard; both are covered by tests, and diff behaviour for non-secret data is unchanged.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated
- [x] No breaking changes (or documented) — `CompareData` signature change is internal; all in-repo callers updated

### DIFF
--- a/cli/internal/secret/secret.go
+++ b/cli/internal/secret/secret.go
@@ -1,6 +1,6 @@
 // Package secret provides String, a self-redacting value type for sensitive
 // data (API tokens, passwords, credentials). Every standard Go surface that
-// prints, formats, logs, or serializes a String returns a redacted form
+// prints, formats, logs, or serializes a String returns a masked form
 // automatically; the real value escapes only through the explicit, greppable
 // Reveal() accessor. Sensitivity becomes a property of the field's type, so a
 // secret is masked once and stays masked at every surface it flows through.
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	redactedPlaceholder = "(remote-redacted)"
-	shortMask           = "***"
+	unknownPlaceholder = "(unknown)"
+	shortMask          = "***"
 	// hintThreshold is the minimum number of characters before masked() reveals
 	// the last four as a spot-check hint; below it the value is fully masked so the
 	// hint never exposes a meaningful fraction of a short secret.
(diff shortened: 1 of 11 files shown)


cli/internal/secret/secret.go
```diff
@@ -29,37 +29,53 @@ const (
...
+// Diff reports whether a and b differ for plan purposes — true means the secret
+// must be (re-)applied. An unknown secret always differs, even from another
+// unknown: backend APIs never return secrets, so we can never confirm the remote
+// matches the local value and must re-apply every run. Two known secrets differ
+// only when their real values differ.
+func (a String) Diff(b String) bool {
+	if a.unknown || b.unknown {
+		return true
+	}
+	return a.v != b.v
```
sivashanmukh, line 76: 🟡 A scenario worth considering for this rule: since the remote is always unknown, `New("").Diff(NewUnknown())` returns `true`, so an *unset* local secret still forces a re-apply every run. If a provider's Update reveals the field without guarding `IsZero()`, that sends `""` and overwrites the backend secret with empty — repeatedly (confirmed against this branch: `New("").IsZero()==true`, `Diff==true`, `Reveal()==""`). Might be worth having `Diff` treat an unset local secret as no-diff (e.g. `if a.IsZero() { return false }` at the top), so an omitted secret means "leave the remote alone" rather than "wipe it" — the trade-off being you can no longer clear a secret by emptying the spec, which is arguably cleaner done explicitly anyway. Not a blocker — flagging for you to weigh.

cli/internal/syncer/differ/diff.go
```diff
@@ -150,42 +199,58 @@ func CompareData(r1, r2 resources.ResourceData) map[string]PropertyDiff {
...
+		// unknown secret always diffs (we can't read the remote, so we re-apply),
+		// otherwise known values are compared. The diff is flagged Secret so the
+		// reporter renders it distinctly and the resource is classified as always
+		// re-applied. SourceValue/TargetValue stay secret.String so they mask. A
+		// *secret.String (the form that survives RawData's struct→map decode, like
+		// *PropertyRef) is dereferenced to this case by the pointer branch above.
+		case secret.String:
+			v2Typed := v2.(secret.String)
+			if v1Typed.Diff(v2Typed) {
+				record(key, PropertyDiff{Property: key, SourceValue: v1, TargetValue: v2, SecretOnly: true})
```
sivashanmukh, line 220: 🟡 Something to consider (we may not hit this in practice): the secret case only fires for a directly-typed `secret.String` (or `*secret.String` via the pointer branch). A secret nested in `[]map[string]any` / `[]any` falls back to `reflect.DeepEqual`, so it isn't flagged `SecretOnly`. Running it against this branch — same `secret.New("…")` vs `NewUnknown()`:

```
typed token  -> SecretOnly=true   -> "Always re-applied" section
nested token -> SecretOnly=false  -> "Updated resources":
    creds: [map[token:****cret]] => [map[token:(unknown)]]
```

So a freeform/slice secret (design edge-case #4, `WrapMap`) would surface as genuine drift every run instead of "Always re-applied." No leak — `Format` still masks under `%v` — but the always-re-apply UX wouldn't cover this shape. Might be worth documenting the limitation, and perhaps a render-path no-leak test for the slice case. Flagging for you to weigh.

cli/internal/syncer/differ/diff.go
```diff
@@ -33,15 +34,48 @@ func (d *Diff) HasDiff() bool {
...
 	Diffs map[string]PropertyDiff
+	// SecretOnly is true when every property diff is secret-driven, so the
+	// resource is "always re-applied" rather than genuine drift. It is computed
+	// once while the diffs are built (see compareData) and cached here.
+	SecretOnly bool
+}
+
+// IsSecretOnly reports whether this resource updates only because of unknown
+// secrets. The verdict is precomputed in ComputeDiff, so this is O(1).
+func (rd ResourceDiff) IsSecretOnly() bool {
```
sivashanmukh, line 66: [nit] `IsSecretOnly()` is a redundant accessor over the exported `SecretOnly` field (tests set `.SecretOnly` directly). Either unexport the field and keep the method, or drop the method and use the field.

sivashanmukh (commented):
Reviewed for leaks and against the base (#608) and the design doc — **no plaintext leak**, masking holds across every render path including container `%v` via `Format`, and the tests assert no value escapes. Solid, well-tested diff plumbing. A few inline notes on edge cases of the always-re-apply policy; none blocking.

cli/internal/secret/secret.go
dhawal1248: Good point, lets include this in the diff revamp for secrets: how do we show this to a user that a secret would overwritten by an empty string

cli/internal/syncer/differ/diff.go
dhawal1248: will document this

cli/internal/secret/secret.go
dhawal1248: Created [DEX-425](https://linear.app/rudderstack/issue/DEX-425/align-on-and-revamp-how-secrets-are-reported-in-apply-cycle-diffs) for this

cli/internal/syncer/differ/diff.go
dhawal1248: done and added a test case too [chore: add doc comment for secrets in slices and a test case to lock …](https://github.com/rudderlabs/rudder-iac/pull/614/commits/535c8b5012e35375c1c67256c665bf3c9c0ea2a4)

sivashanmukh (approved):

PR #616, opened 2026-06-10
feat(import): scaffold secrets as variable references with a companion var file (DEX-410)

dhawal1248:
## 🔗 Ticket

Resolves [DEX-410](https://linear.app/rudderstack/issue/DEX-410/import-scaffolding-for-secrets-via-variable-substitution)

---

## Summary

Importing a resource with secret fields now produces a usable spec plus a fill-in-the-blanks var file, composing with variable substitution. A remote never returns a secret's real value, so exporting it would only yield a masked literal that can't be applied; instead, each secret field is emitted as a quoted `{{ .VAR }}` reference and the importer scaffolds `imported/secrets.vars.yaml` with one placeholder per variable. On a subsequent `apply --var-file`, the existing substitution chain injects the real value, which loads back into `secret.String`. Gated on the `enableVarSubstitution` experimental flag. Stacked on #614 (DEX-408).

---

## Changes

- `secret.String` gains an optional substitution variable name, attached via the `secret.WithVariableName(...)` constructor option. When set, `MarshalJSON`/`MarshalYAML` emit a quoted `{{ .VAR }}` reference instead of the masked literal, so the token flows through the untouched export pipeline (`ToMap` → YAML formatter) with no framework changes; every other surface (fmt, logs) keeps redacting
- Providers name their variables explicitly in `MapRemoteToSpec` (e.g. `BOOK_<externalID>_ACCESS_KEY`), keeping names deterministic and stable across re-imports; `WithVariableName` uses the name verbatim (grammar compliance is the provider's responsibility) and is the single place the experimental gate is checked (gate off → name dropped → masked literal, the pre-scaffolding behavior)
- Importer scans generated YAML entities for variable references (`varsubst.ExtractVariableNames`) and scaffolds `imported/secrets.vars.yaml` via the writer/formatter framework, merging with a pre-existing var file so user-filled values are never overwritten; placeholders are unfilled (`KEY:`), which the var-file resolver rejects, so a forgotten secret fails apply loudly instead of sending `""`
- Spec loader skips `*.vars.yaml` / `*.vars.yml` so var files inside the project directory are not parsed as specs (documented in the varsubst README)
- `secret.String` implements mapstructure v2.5's `Unmarshaler` interface (`UnmarshalMapstructure`), so substituted real values load back into `secret.String`/`*secret.String` through any plain `mapstructure.Decode` (the variable name is cleared on every unmarshal — it only exists on the export path) — no per-decoder hook wiring in `BaseHandler` or providers' hand-rolled `LoadSpec` decoders (dep bump: `go-viper/mapstructure` v2.4.0 → v2.5.0)
- Example test provider: `AccessKey` secret on books (spec carries `secret.String`, resource carries `*secret.String` per the differ's pointer convention), workspace ID in remote metadata, and `Import` now sets the external ID (was a stub passing `""`)

---

## Testing

- Unit tests: variable-name marshaling on `secret.String` (token emission, gate on/off, no-leak across fmt/json surfaces, verbatim name use, unmarshal clears the name); type-level mapstructure decoding for all secret field shapes; `varsubst.ExtractVariableNames`; var-file scaffold/merge/no-op/gate-off; loader var-file skip
- Integration test (`provider_import_test.go`): full cycle against the in-memory example backend — import yields a spec with a quoted variable reference (no mask) and a var file with a null placeholder; applying unfilled is rejected; after filling, the first apply imports/links and the second re-sends the secret, round-tripping the real value to the backend
- `make lint` clean, `make test` (race) green
- Reviewed by Codex; high findings (silent empty-secret apply) fixed; an earlier framework-side reflection walk that derived variable names was replaced with explicit provider-named variables (now the `WithVariableName` option on `secret.String`)

---

## Risk / Impact

Low
Notes: all new behavior is behind the `enableVarSubstitution` experimental flag; with the gate off, export output is unchanged (masked literals). The loader newly skips `*.vars.yaml` files, which previously failed loading as invalid specs — strictly a relaxation.

---

## Deviations from the ticket

- **"Collision-free" variable names**: the ticket asked for framework-guaranteed collision-free names. The final design deliberately delegates naming to the provider (`WithVariableName`), so collision-freedom is a provider convention (names built from resource type + external ID), not a framework invariant. A framework-side name-derivation walk with collision detection was implemented first and then removed by design review: explicit names are simpler, greppable, and let a provider intentionally share one variable across fields. Reusing a name across distinct secrets is therefore possible and is the provider author's responsibility to avoid.
- **Placeholder semantics** (open question in the design doc): placeholders are scaffolded as unfilled YAML nulls (`KEY:`) rather than `KEY: ""`, so applying with a forgotten secret fails loudly at var-file load instead of silently sending an empty value.
- **New project convention**: var files inside the project directory must be named `<name>.vars.yaml`/`.vars.yml` — every other YAML in the project is strict-parsed as a spec. Documented in the varsubst README.
- **Re-import merge**: the var-file merge path is unreachable via `rudder-cli import` today (the command refuses an existing `imported/` directory); it exists for the intended workflow where specs are moved out of `imported/` but the var file is kept, and for any future relaxation of that guard.
- **Dependency bump**: `go-viper/mapstructure` v2.4.0 → v2.5.0, for the type-level `Unmarshaler` interface (also refreshes stale `// indirect` annotations in `go.mod`, e.g. `participle`).

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/project/importer/importer.go
```diff
@@ -89,10 +89,20 @@ func WorkspaceImport(
...
+	if err := writer.Write(ctx, importDir, formatters, entities); err != nil {
 		return fmt.Errorf("writing files for formattable entities: %w", err)
 	}
 
+	varFile, err := scaffoldSecretsVarFile(importDir, entities)
+	if err != nil {
+		return fmt.Errorf("scaffolding secrets var file: %w", err)
+	}
+	if varFile != "" {
+		fmt.Printf("Imported specs reference variables for secret values.\n"+
```
dhawal1248: CLAUDE: we usually dont use fmt.Printf to log user messages..we use the ui library

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,167 @@
...
+`
+
+// scaffoldSecretsVarFile writes a fill-in-the-blanks var file for every
+// variable referenced by the generated entities. Only active under the
+// enableVarSubstitution experimental gate — without substitution the
+// references could never be resolved on apply. If a var file already exists
+// (e.g. kept from an earlier import), values the user filled in are kept and
+// only missing variables gain a placeholder.
+func scaffoldSecretsVarFile(baseDir string, entities []writer.FormattableEntity) (string, error) {
+	if !config.GetConfig().ExperimentalFlags.EnableVarSubstitution {
```
dhawal1248, line 37: this experimental flag will be removed in the next PR

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,167 @@
...
+
+	for _, name := range names {
+		if _, ok := vars[name]; !ok {
+			// nil scaffolds a "KEY:" line; the var-file resolver rejects null
+			// values, so an unfilled placeholder fails apply loudly.
+			vars[name] = nil
+		}
+	}
+
+	data, err := renderVarFile(vars)
```
dhawal1248: CLAUDE: check if we can leverage the framework functions to create files.All providers create specs and then the framework actually creates those..there are other places where the file is created and written, reuse that functionality.

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// 0600: the user is expected to put real secrets in this file.
+	if err := os.WriteFile(path, data, 0600); err != nil {
+		return "", fmt.Errorf("writing var file %s: %w", path, err)
+	}
+
+	return path, nil
+}
+
+func loadExistingVars(path string) (map[string]any, error) {
```
dhawal1248: explain what this func does

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,167 @@
...
+		return "", nil
+	}
+
+	names := collectVariableNames(entities)
+	if len(names) == 0 {
+		return "", nil
+	}
+
+	path := filepath.Join(baseDir, SecretsVarFileName)
+	vars, err := loadExistingVars(path)
```
dhawal1248: CLAUDE: clarify that if a pre existing var file exists,we merge any new variables that are discovered during import with the existing file

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,167 @@
...
+
+	names := make([]string, 0, len(found))
+	for name := range found {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return names
+}
+
+func extractVariableNames(content any, found map[string]struct{}) {
```
dhawal1248, line 127: CLAUDE: explain what this does

dhawal1248 (commented):
Project package review comments

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+			// values, so an unfilled placeholder fails apply loudly.
+			vars[name] = nil
+		}
+	}
+
+	entity := writer.FormattableEntity{
+		Content:      vars,
+		RelativePath: path,
+	}
+	if err := writer.OverwriteFile(formatter.Setup(varFileFormatter{}), entity); err != nil {
```
sivashanmukh: writer.OverwriteFile apparently creates the file in 644 which is world readable while your tests everywhere rightly use 600. 

write the var file with 0600 (a dedicated os.WriteFile(path, formatted, 0o600) here, or an optional mode on the writer entity. A mode may be better to ensure that this option is available for the next dev using this to consciously use the right mode.

cli/internal/provider/testutils/example/handlers/book/handler.go
```diff
@@ -186,6 +198,19 @@ func (h *HandlerImpl) Delete(ctx context.Context, id string, oldData *model.Book
 	return h.backend.DeleteBook(oldState.ID)
 }
 
+func accessKeyVarName(externalID string) string {
+	return fmt.Sprintf("BOOK_%s_ACCESS_KEY", strings.ToUpper(strings.ReplaceAll(externalID, "-", "_")))
+}
+
+// revealAccessKey is the single point where the real secret escapes toward the
+// backend API; a spec without an access key sends the empty string.
+func revealAccessKey(data *model.BookResource) string {
```
sivashanmukh, line 207: 🟡 `revealAccessKey` returns `""` for a nil/unset secret and `Update` sends it unguarded. Combined with #614's always-re-apply rule (`New("").Diff(NewUnknown()) == true`), a spec that simply omits `accessKey` re-applies on every run and overwrites the backend's stored secret with `""` — repeatedly. This is exactly the scenario tracked in [DEX-425](https://linear.app/rudderstack/issue/DEX-425/align-on-and-revamp-how-secrets-are-reported-in-apply-cycle-diffs), but the example provider is the adoption template, so DEX-409 adopters will copy this shape verbatim.

Suggest either an `IsZero()` guard here (skip/preserve on unset), or — if the example should stay minimal — a loud comment on `revealAccessKey` saying real providers must decide their unset-secret policy at this point, linking DEX-425.

sivashanmukh (commented):
Three follow-ups from the stack review (security lens). None of them blocks the framework itself; the first one matters most because this example is the adoption template.

cli/internal/secret/secret.go
```diff
@@ -35,17 +37,53 @@ type String struct {
...
+// the export spec chooses the name from what it knows about the resource
+// (e.g. resource type, external ID, field) — it is used verbatim, so it must
+// satisfy the substitutor's variable grammar (^[A-Za-z_][A-Za-z0-9_]*$) —
+// which is what keeps names deterministic and stable across re-imports.
+//
+// Scaffolding only works under the enableVarSubstitution experimental gate —
+// without substitution the reference could never be resolved on apply — so
+// with the gate off this option is a no-op and the secret exports as a masked
+// literal, the pre-scaffolding behaviour.
+func WithVariableName(name string) Option {
```
sivashanmukh, line 65: 🟡 Verbatim-by-design is fine, and I confirmed the failure is at least loud: the substitutor rejects a malformed token at apply (`ErrInvalidVarSyntax`). But that failure surfaces two steps away from the bug: import "succeeds", `ExtractVariableNames` skips the malformed token so **no placeholder is scaffolded**, and the user discovers it only when apply fails on a spec they didn't write. Since variable names are derived from external IDs (user-controlled resource names upstream), validating here — or erroring at import time when a generated spec contains a token extraction can't parse — would move the failure to the moment it's created. Cheap insurance: the grammar regex (`validVarName`) already exists in `varsubst`.

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+		return "", fmt.Errorf("writing var file: %w", err)
+	}
+
+	return path, nil
+}
+
+// loadExistingVars reads the var file at path into a name→value map so a new
+// scaffold can merge with it. A missing file is not an error — it simply means
+// there is nothing to merge and every variable starts as a placeholder.
+func loadExistingVars(path string) (map[string]any, error) {
```
sivashanmukh: nit: the merge path round-trips the existing file through `yaml.Unmarshal` into a map, so any comments the user added are lost on rewrite. Acknowledged this path is unreachable today (import refuses an existing `imported/`); just worth a line in the doc comment so whoever relaxes that guard knows.

cli/internal/project/importer/importer.go
```diff
@@ -89,10 +90,20 @@ func WorkspaceImport(
...
 	}
 
+	varFile, err := scaffoldSecretsVarFile(importDir, entities)
+	if err != nil {
+		return fmt.Errorf("scaffolding secrets var file: %w", err)
+	}
+	if varFile != "" {
+		ui.PrintWarning(fmt.Sprintf("Imported specs reference variables for secret values.\n"+
+			"Fill in the placeholders in %s (keep it out of version control) and pass it to apply via --var-file.", varFile))
+	}
```
abhimanyubabbar: Why should it be a warning ?

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+
+	entity := writer.FormattableEntity{
+		Content:      vars,
+		RelativePath: path,
+	}
+	if err := writer.OverwriteFile(formatter.Setup(varFileFormatter{}), entity); err != nil {
+		return "", fmt.Errorf("writing var file: %w", err)
+	}
+
+	return path, nil
```
abhimanyubabbar: Should we just try to create a var's file and not overwrite it ?

If it fails on writing ( which shouldn't happen ) we just error out saying a file matching secrets.vars.yaml exists and error out. 

Wdyt ?

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+	if len(names) == 0 {
+		return "", nil
+	}
+
+	path := filepath.Join(baseDir, SecretsVarFileName)
+
+	// If a var file already exists (e.g. kept from an earlier import), merge
+	// into it: variables discovered by this import gain a placeholder only when
+	// missing, so values the user already filled in are never overwritten.
+	vars, err := loadExistingVars(path)
```
abhimanyubabbar: This should not be our responsibility of merging. If we have earlier file at location we should simply error out.

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+// provider or strategy produced them. Only YAML entities count: variable
+// substitution runs exclusively over spec files, so references in other
+// generated files (e.g. extracted SQL or code) could never be resolved.
+func collectVariableNames(entities []writer.FormattableEntity) []string {
+	found := make(map[string]struct{})
+	for _, entity := range entities {
+		ext := filepath.Ext(entity.RelativePath)
+		if ext != loader.ExtensionYAML && ext != loader.ExtensionYML {
+			continue
+		}
```
abhimanyubabbar: When would we have entities in a path which is not YAML or extensionYML ?

cli/internal/project/importer/varfile.go
```diff
@@ -0,0 +1,182 @@
...
+	case []any:
+		for _, item := range v {
+			extractVariableNames(item, found)
+		}
+	case string:
+		for _, name := range varsubst.ExtractVariableNames([]byte(v)) {
+			found[name] = struct{}{}
+		}
+	}
+}
```
abhimanyubabbar, line 148: Although having generic branches help, but curious when would we be hitting any other branch factor like map[string]any etc ?

cli/internal/secret/secret.go
```diff
@@ -35,17 +37,53 @@ type String struct {
...
+	return func(s *String) {
+		if !config.GetConfig().ExperimentalFlags.EnableVarSubstitution {
+			return
+		}
```
abhimanyubabbar, line 69: Maybe we don't need this at this layer. It might be too much of defensive check

cli/internal/secret/secret.go
```diff
@@ -35,17 +37,53 @@ type String struct {
...
+		s.varName = name
+	}
 }
 
 // New wraps a real, known value. The spec loader and provider spec-to-args
 // conversion use it.
-func New(v string) String { return String{v: v} }
+func New(v string, opts ...Option) String { return apply(String{v: v}, opts) }
```
abhimanyubabbar, line 76: Would we ever create the New with opts in the caller identity ?

cli/internal/secret/secret.go
```diff
@@ -35,17 +37,53 @@ type String struct {
...
+// "{{ .name }}" reference instead of a masked literal. The provider building
+// the export spec chooses the name from what it knows about the resource
+// (e.g. resource type, external ID, field) — it is used verbatim, so it must
+// satisfy the substitutor's variable grammar (^[A-Za-z_][A-Za-z0-9_]*$) —
+// which is what keeps names deterministic and stable across re-imports.
+//
+// Scaffolding only works under the enableVarSubstitution experimental gate —
+// without substitution the reference could never be resolved on apply — so
+// with the gate off this option is a no-op and the secret exports as a masked
+// literal, the pre-scaffolding behaviour.
```
abhimanyubabbar, line 64: Can you simplify this comment ?
It was very confusing for me to understand

cli/internal/varsubst/utils_test.go
```diff
@@ -0,0 +1,51 @@
...
+		want []string
+	}{
+		{
+			name: "no tokens",
+			in:   "accessKey: plain-value",
+			want: []string{},
+		},
+		{
+			name: "single token",
+			in:   `accessKey: "{{ .ACCESS_KEY }}"`,
```
abhimanyubabbar, line 22: If I am not mistaken, we wanted to have input as `accessKey: {{ .ACCESS_KEY }}` without the quotes right ?

cli/internal/project/importer/importer.go
dhawal1248: DO: change to info

cli/internal/project/importer/varfile.go
dhawal1248: DO: remove the merging logic of var file and switch to WriteFile instead of OverwriteFile..

cli/internal/project/importer/varfile.go
dhawal1248: DO: remove the yaml check

cli/internal/project/importer/varfile.go
dhawal1248: DO: doublecheck this

cli/internal/varsubst/utils_test.go
dhawal1248: DO: doublecheck - this should be generated without quotes

abhimanyubabbar (approved):

cli/internal/provider/testutils/example/handlers/book/handler.go
dhawal1248: It is simplest to pass on the empty value as of now.Even if we decide whether or not to send the field in the API request based on the `IsZero()` func, we do not how the API would interpret it, an empty value could either mean drop the secret/leave it as it.

PR #623, opened 2026-06-15
chore: add e2e test for variable susbstitution

dhawal1248:
## 🔗 Ticket

Resolves [DEX-428](https://linear.app/rudderstack/issue/DEX-428)

---

## Summary

Extends the existing apply/migrate E2E suite with end-to-end coverage for the experimental variable-substitution feature, and wires variable substitution into the `migrate` command so that projects which use `{{ .VAR }}` placeholders can be migrated.

---

# Description

### 🆕 Features

- **`migrate` now resolves `{{ .VAR }}` placeholders.** `cli/internal/cmd/project/migrate/migrate.go` builds the project via `app.NewProjectOptions(config.GetConfig(), varFiles)` and adds a `--var-file` flag, so placeholders are resolved (from `RUDDER_*` env vars / var files) before the project is validated and rewritten to v1. Gated by the same `enableVarSubstitution` experimental flag as `apply`/`validate` — a no-op when the flag is off.
- **E2E coverage for substitution + precedence.** `TestProjectApply` (`cli/tests/command_apply_test.go`) enables the experimental flags and exercises two placeholders on the `api_tracking` event across the create, update, and post-migration apply flows:
  - `API_TRACKING_DESCRIPTION` — resolved from the var file only.
  - `API_TRACKING_NAME` — set in both the var file **and** `RUDDER_API_TRACKING_NAME`; the env var wins, pinning the documented precedence (env > var file). The var file value is deliberately wrong, so a precedence regression would fail the snapshot.

### 🏗 Architecture Changes

- `migrate` now goes through the same project-options assembly (`NewProjectOptions` → `WithSubstitutor`) as `apply`/`validate` instead of constructing a bare project. The migrator emits `project.Specs()` (the loaded, **substituted** specs), so migrated v1 output contains the resolved values.
- The experimental project loader (`cli/pkg/exp/project.Load`) is likewise made substitution-aware via `NewProjectOptions`, so it behaves consistently with the real commands (no-op when the flag is off).

### 📦 Code Organization / Tests

- `--var-file` is threaded through `applyAndVerify`, `verifyNoChangesToApply`, and `copyAndMigrateProject` so apply, dry-run, and migrate all resolve the placeholders consistently.
- The placeholders live in the shared `testdata/project/create/events.yaml` and `update/events.yaml` fixtures; a new flat `testdata/project/substitution.vars.yaml` supplies the values. Resolved values equal the previously committed values, so **no upstream snapshots change**.

### 🔄 Migration

- Migrating a substitution-using project resolves and bakes in the values during migration. Behaviour is unchanged when the experimental flag is off.

---

## Why wire the substitutor into the migrate project loader?

`migrate` loads and validates the v0.1 project in `PreRunE` (`proj.Load(location)`) **before** rewriting it to v1. Without a substitutor in that loader, the placeholders are validated as literal tokens, which breaks in two ways:

1. **Field-content rules reject the literal token.** An event `description`, for example, must start with a letter (`datacatalog/events/spec-syntax-valid`), so a literal `{{ .API_TRACKING_DESCRIPTION }}` fails validation and migration aborts.
2. **File/env-backed values are never supplied.** A placeholder meant to come from a `--var-file` or `RUDDER_*` env var has no value during migrate.

The net effect was that a project which had adopted variable substitution could not be migrated at all. Wiring the same substitutor `apply`/`validate` use makes `migrate` resolve placeholders before validation and rewrite, keeping the three commands consistent. Because it is gated behind the experimental flag, projects that don't use the feature are unaffected.

---

## Testing

- `make build`, `make lint` (0 issues), and `make test` (unit) pass.
- Verified offline with the built binary: `validate` of the create/update specs resolves correctly, and the full `migrate --var-file` round-trip substitutes (`name` from the env var wins, `description` from the var file) with the migrated specs re-validating.
- Note: a full live `make test-e2e` requires a clean workspace — the suite's 2-minute per-command timeout is exceeded by `destroy` when the target workspace already holds a large number of resources (unrelated to this change; `destroy` does not use substitution).

---

## Risk / Impact

**Low.** All new behaviour is gated behind the `enableVarSubstitution` experimental flag; with the flag off, `migrate`/`apply`/`validate` behave exactly as before. No production code paths change for non-substitution projects, and no snapshots were modified.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated
- [x] No breaking changes (gated behind experimental flag)

### DIFF
--- a/cli/internal/cmd/project/apply/apply.go
+++ b/cli/internal/cmd/project/apply/apply.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/rudderlabs/rudder-iac/api/client"
 	"github.com/rudderlabs/rudder-iac/cli/internal/app"
 	"github.com/rudderlabs/rudder-iac/cli/internal/cmd/telemetry"
 	"github.com/rudderlabs/rudder-iac/cli/internal/config"
@@ -24,13 +25,14 @@ var (
 
 func NewCmdApply() *cobra.Command {
 	var (
-		deps     app.Deps
-		p        project.Project
-		err      error
-		location string
-		dryRun   bool
-		confirm  bool
-		varFiles []string
+		deps      app.Deps
+		p         project.Project
+		workspace *client.Workspace
+		err       error
+		location  string
+		dryRun    bool
+		confirm   bool
+		varFiles  []string
 	)
 
 	cmd := &cobra.Command{
@@ -52,10 +54,16 @@ func NewCmdApply() *cobra.Command {
 				return fmt.Errorf("initialising dependencies: %w", err)
 			}
 
+			workspace, err = deps.Client().Workspaces.GetByAuthToken(context.Background())
+			if err != nil {
+				return fmt.Errorf("fetching workspace information: %w", err)
+			}
+
 			projectOpts, err := app.NewProjectOptions(config.GetConfig(), varFiles)
 			if err != nil {
 				return err
 			}
+			projectOpts = append(projectOpts, project.WithWorkspaceID(workspace.ID))
 
 			p = deps.NewProject(projectOpts...)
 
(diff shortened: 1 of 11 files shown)


abhimanyubabbar (approved):
Changes look good to start with.

PR #627, opened 2026-06-22
feat: add import-manifest syntactic validation rules

Jayachand:
## 🔗 Ticket

resolves [DEX-449](https://linear.app/rudderstack/issue/DEX-449)

---

## Summary

PR 3 of the import-manifest plan. Wires the import-manifest provider into the validation registry and gives it syntactic rules, so a manifest file is now a known, structurally-validated kind. `BuildRegistry` becomes two-provider (resource + manifest): it unions both providers' match patterns (so `spec-syntax-valid` accepts `import-manifest`) and registers the manifest provider's `SyntacticRules()`. The provider's `ParseSpec` is implemented for real (URN + JSON-pointer extraction), replacing the skeleton stub. This is syntactic-only — the `SemanticRules()` seam (orphaned-URN) lands in a later PR with the rule that fills it.

**Sample errors:**

```
error[import-manifest/spec-syntax-valid]: remote_id is required
  --> cli/pkg/exp-import/import-manifest-2.yaml:154:11
       |
   154 | remote_id:
       | ^^^^^^^^^^^


error[import-manifest/duplicate-urn]: duplicate URN 'tracking-plan:api-tracking' in workspace '2KDL30oD9PHzkaEi8PLOfodqInA'
  --> cli/pkg/exp-import/import-manifest.yaml:153:11
       |
   153 | urn: tracking-plan:api-tracking
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

---

## Changes

* `project.BuildRegistry(provider, manifestProvider)` — union match patterns into the active set; append `manifestProvider.SyntacticRules()`. **No** `SemanticRules()` append yet (no rule to fill it). Threaded the new arg through `registry()`, `ruledoc.Build`, and `app.deps.Registry`.
* `importmanifest.Provider.ParseSpec` — real URN/JSON-pointer extraction (was a stub).
* New rule `import-manifest/workspaces-valid` (per-file): workspaces present, `workspace_id` required, urn-based resources (`local_id` rejected), no intra-workspace duplicate urn — keyed by `(workspace_id, urn)`.
* New rule `import-manifest/duplicate-urn` (cross-file `MultiSpecRule`): same `(workspace_id, urn)` across more than one manifest file.
* Two `*.docs.yaml` rule-doc fragments + `RuleDocEntries()` wiring; `ruledoc.Build` now collects the manifest provider's fragments.
* New leaf package `manifestspec` (kind constant + workspaces decode) so the provider and its rules avoid an import cycle, matching the existing per-provider pattern.
* Test-provider fixups: gave the `example` and several `project` mock providers explicit `SupportedMatchPatterns` so the now-active known-kind gate treats their fixture kinds as known (these tests previously relied on an empty pattern set disabling the gate).

---

## Testing

* Unit (TDD, red→green for each new function): `BuildRegistry` aggregation, `ParseSpec` extraction, both rules (table-driven, full `[]ValidationResult` comparison), provider rule-set wiring.
* Rule-doc generator tests execute the new fragments' examples and assert the exact diagnostics (`internal/ruledoc`, `internal/app`).
* Full unit suite green (`make test`); `make lint` 0 issues. E2E (`cli/tests`) not run — unaffected by this change.

---

## Risk / Impact

Medium — touches the shared `BuildRegistry` signature (all callers updated, compiler-enforced) and activates the known-kind gate for the manifest pattern, which surfaced (and fixed) latent test-provider assumptions. Manifest validation is new but additive; resource validation is unchanged.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


cli/internal/project/importmanifest/rules/manifest_spec_valid.go
```diff
@@ -0,0 +1,97 @@
...
+func (r *manifestSpecValidRule) Examples() vrules.Examples { return vrules.Examples{} }
+
+func (r *manifestSpecValidRule) Validate(ctx *vrules.ValidationContext) []vrules.ValidationResult {
+	workspaces, err := manifestspec.DecodeWorkspaces(ctx.Spec)
+	if err != nil {
+		return []vrules.ValidationResult{{Reference: "/spec", Message: "spec.workspaces is malformed"}}
+	}
+	if len(workspaces) == 0 {
+		return []vrules.ValidationResult{{Reference: "/spec/workspaces", Message: "manifest has no workspaces"}}
+	}
```
abhimanyubabbar, line 48: Should this be an error ?

cli/internal/project/importmanifest/rules/manifest_spec_valid.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var results []vrules.ValidationResult
+	for wi, ws := range workspaces {
```
abhimanyubabbar: Is wi index ?
If yes, should we name it better ?

abhimanyubabbar (approved):
Approving the PR with minor comments 👍🏼

cli/internal/project/importmanifest/rules/manifest_spec_valid.go
Jayachand: it is workspaceIndex and workspace.

cli/internal/project/importmanifest/rules/manifest_spec_valid.go
Jayachand: It is the intent. User should not declare manifest yaml without any entries.
Do you think it is more restricitve and should relax it ?

Jayachand: We would be going with this restriction for now. Can relax if it is too restrictive

cli/internal/project/importmanifest/rules/manifest_spec_valid.go
Jayachand: renamed to (i, workspace) (j, resource) for clarity

PR #628, opened 2026-06-22
feat(typer): add --local flag to generate from local specs

sivashanmukh:
## 🔗 Ticket

Resolves [DEX-450](https://linear.app/rudderstack/issue/DEX-450/generate-data-gov-types-via-rudder-cli-without-control-plane)

---

## Summary

`typer generate` can only source a tracking plan from the **remote** workspace, so to generate code reflecting local tracking-plan edits a user must `apply` first — coupling code generation to deployment and requiring auth/network.

This adds `typer generate --local`, which builds the tracking plan from the **local YAML specs** (the same artifacts `import` emits and `apply` consumes), skipping the remote fetch. Users can iterate on their tracking plan and generated code together, offline, without applying.

It reuses the existing `PlanProvider` seam: a second implementation renders the loaded local catalog into the same JSON-Schema representation the remote returns and feeds it through a single shared parser (`BuildTrackingPlan`), so **local and remote output stay identical**.

---

## Changes

- Extract `BuildTrackingPlan` in `jsonschema.go` as the shared parser entry point (remote provider delegates to it; no behavior change)
- Add `LocalCatalogPlanProvider` + renderer (`local.go`): primitives, multi-type, enum, arrays, nested objects, and custom types → `$defs`/`$ref` (with a cycle guard); resolves rule refs via the existing `ExpandRefs`
- Fail loud on variants (rule- and custom-type-level), deferred to a follow-up
- Wire `--local` and `-l/--location` into `typer generate`; load + validate the project offline (no auth/network) and select the tracking plan (single-plan default, else error listing ids)

---

## Testing

- **Unit** (`local_test.go`): full-struct assertion of the produced `plan.TrackingPlan` for a primitives/enum/array/nested/custom-type catalog; fail-loud-on-variants and not-found cases
- **Manual**: built the CLI and ran `typer generate --local` against a local project — generates valid TypeScript with correct required/optional fields, no auth; the variant-containing fixture fails loud with a clear message
- `make lint` clean; touched-package tests green

---

## Risk / Impact

Low. New behavior is gated behind `--local` and the stable `PlanProvider` interface; the remote path and the generators are unchanged. The shared-parser refactor is covered by the existing parser tests.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/typer/plan/providers/local.go
```diff
@@ -0,0 +1,457 @@
...
+	ct := p.dc.CustomType(m[1])
+	if ct == nil {
+		return nil, false, fmt.Errorf("custom type %q not found in local specs", m[1])
+	}
+
+	if existing, ok := reg[ct.Name]; ok {
+		return existing, true, nil
+	}
+
+	pct := &plan.CustomType{Name: ct.Name, Description: ct.Description}
```
akashrpo: Local sets Description on the custom type; [remote](https://github.com/rudderlabs/rudder-iac/blob/main/cli/internal/typer/plan/providers/jsonschema.go#L198) never does. Both generators render it as a doc comment, so the same custom type yields different output local vs remote.

cli/internal/typer/plan/providers/local.go
```diff
@@ -0,0 +1,457 @@
...
+	property := plan.Property{Name: in.Name}
+
+	if ct, ok, err := p.customType(in.Type, reg); err != nil {
+		return property, nil, err
+	} else if ok {
+		property.Types = []plan.PropertyType{ct}
+		return property, nil, nil
+	}
+
+	if len(in.Types) > 0 {
```
akashrpo, line 197: It returns type and config only - never reads ItemTypes or builds nested object schema unlike remote's [parseTypeDefinition](https://github.com/rudderlabs/rudder-iac/blob/main/cli/internal/typer/plan/providers/jsonschema.go#L253). A multi-type array/object property generates a less-precise type under --local.

cli/internal/typer/plan/providers/local.go
```diff
@@ -0,0 +1,457 @@
...
+	primType, err := parsePrimitiveType(ct.Type)
+	if err != nil {
+		return nil, false, fmt.Errorf("custom type %q: %w", ct.Name, err)
+	}
+	pct.Type = primType
+	pct.Config = toConfig(ct.Config)
+
+	switch ct.Type {
+	case "object":
+		schema := &plan.ObjectSchema{Properties: map[string]plan.PropertySchema{}, AdditionalProperties: true}
```
akashrpo: Local hardcodes AdditionalProperties: true for object custom types; remote derives it
cc @fxenik

akashrpo (changes requested):

cli/internal/typer/plan/providers/local.go
sivashanmukh: Fixed in 00e4f319. Dropped `Description` from the local custom type so it matches the remote parser (which never sets it). Converging on remote is the spec's contract (remote path untouched); if we want the custom-type description in doc comments that's a separate change to both paths, including the remote JSON-Schema parser.

cli/internal/typer/plan/providers/local.go
sivashanmukh: Fixed in 00e4f319. The multi-type branch now resolves array `ItemTypes` and nested object schema off the first type, mirroring remote's `parseTypeDefinition` (which keys nested/array handling off `Types[0]`). Covered by TestLocalCatalogPlanProvider_MatchesRemoteConventions.

cli/internal/typer/plan/providers/local.go
sivashanmukh: Fixed in 00e4f319. Object custom types now derive `AdditionalProperties` from the type's config (`additional_properties`/`additionalProperties`), defaulting to true — matching the remote parser's default for absent `additionalProperties`. An explicit `false` in config is now honored instead of silently forced true. Covered by the new test.

cli/internal/cmd/typer/typer.go
```diff
@@ -61,29 +63,30 @@ func newCmdGenerate() *cobra.Command {
 				return fmt.Errorf("unsupported platform: %s (supported platforms: %s)", platform, strings.Join(supported, ", "))
 			}
 
+			if local && !config.GetConfig().ExperimentalFlags.LocalTyper {
+				return fmt.Errorf("--local is experimental; enable it by setting the 'localTyper' flag (e.g. RUDDERSTACK_X_LOCAL_TYPER=true)")
```
akashrpo, line 67: Nit: the error message tells user to set RUDDERSTACK_X_LOCAL_TYPER=true, but that alone doesn't work (also needs RUDDERSTACK_CLI_EXPERIMENTAL=true)

akashrpo (approved):

PR #632, opened 2026-06-23
feat: broadcast import-manifest into resource providers via consumer interface

Jayachand:
## 🔗 Ticket

resolves [DEX-367](https://linear.app/rudderstack/issue/DEX-367/import-manifest-broadcast-consumer-interface-providerhandler-fan-out)

---

## Summary

Makes `apply`/`validate` consume a central `import-manifest.yaml`. After manifest specs are parsed and aggregated by the project-level import-manifest provider, the combined, workspace-scoped `WorkspacesImportMetadata` is broadcast into the resource provider tree during validation — before graph construction — so handlers attach `ImportMetadata` from the same maps that inline `metadata.import` blocks populate today. Resources matching a manifest URN reconcile to Import (not Create). The broadcast runs after inline loading, so the manifest wins on URN overlap (last writer). This is the read-path half of the import-manifest feature; it does not yet change the write/export path (deferred to a later PR).

---

## Changes

* **`ImportManifestConsumer` interface** (`provider.go`) — optional, nil-safe, deliberately kept out of the master `Provider` interface (type-asserted at the call site) so providers that don't import are not forced to adopt it.
* **`Handler.LoadImportMetadata` promotion** (`baseprovider.go`) — added to the `Handler` interface; `BaseHandler` already implements it, so datagraph/transformations are covered with no per-handler edits.
* **`BaseProvider.LoadImportManifest`** — fans the manifest out to every handler.
* **`CompositeProvider.LoadImportManifest`** — fans out to sub-providers implementing the consumer interface; silently skips the rest.
* **datacatalog** — custom `LoadImportManifest` writing fully-qualified URNs straight into `dc.ImportMetadata` (no kind translation).
* **retl** — extracted reusable `Handler.LoadImportMetadata`; inline path now delegates to it; provider-level fan-out added.
* **event-stream** — extracted `sourceResource.applyImportManifest`; inline path delegates; `Handler.LoadImportMetadata` replicates the manifest into every source (each reads only its own URN at graph time); provider-level fan-out added.
* **`ImportManifest(activeWorkspaceID)` accessor** (`importmanifest/provider.go`) — workspace-scoped: nil when empty, `""`→all (validate), otherwise filtered to the active workspace; nil when no match.
* **Broadcast wired in `project.go`** `handleValidation`, after the loadSpec loop and before `ResourceGraph()`, using `p.workspaceID`.

---

## Testing

* **Unit** (all green via `make test`, `make lint` 0 issues):
  * Accessor scoping (nil-when-empty, `""`→all, workspace filter, no-match→nil).
  * `BaseProvider` / `CompositeProvider` fan-out: nil-safe, broadcasts to consumers, skips non-consumers.
  * datacatalog / retl / event-stream: manifest entries populate the same import-metadata state that inline `metadata.import` does; existing inline-import tests still pass (refactors are behavior-preserving).
  * `project`: end-to-end broadcast through `Load` confirms only the active workspace's entries reach the consumer.

---

## Risk / Impact

Low. New behavior is additive and only triggers when an `import-manifest.yaml` is present and a provider implements the consumer interface; the inline `metadata.import` path is unchanged (the per-provider refactors route both paths through one shared loader). The `Handler` interface gained `LoadImportMetadata`, but every implementer (`BaseHandler` and the custom retl/event-stream handlers) already satisfies it.

> Note: for a hand-written manifest to pass syntax validation on a project whose resource provider declares match patterns, the two-provider registry (separate PR / DEX-449) is required; this PR's broadcast machinery is complete and unit-tested independently.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


cli/internal/providers/datacatalog/localcatalog/catalog.go
```diff
@@ -426,6 +426,28 @@ func addImportMetadata(s *specs.Spec, dc *DataCatalog) error {
...
+	}
+	for _, workspace := range m.Workspaces {
+		for _, resource := range workspace.Resources {
+			if resource.URN == "" {
+				return fmt.Errorf("import-manifest entry in workspace %q is missing urn", workspace.WorkspaceID)
+			}
+			dc.ImportMetadata[resource.URN] = &WorkspaceRemoteIDMapping{
+				WorkspaceID: workspace.WorkspaceID,
+				RemoteID:    resource.RemoteID,
+			}
```
abhimanyubabbar: Given we allow the same URN to be associated to different workspace ID and remote ID, this logic would override the URN for each workspace right ?

cli/internal/provider/provider.go
```diff
@@ -178,6 +178,17 @@ type ConsolidateSyncer interface {
...
 }
 
+// ImportManifestConsumer receives the aggregated, workspace-scoped import
+// metadata after all manifest specs have been loaded. It is implemented by
+// resource providers that populate their internal import-metadata state from a
+// central import-manifest file (see docs/import-manifest-hld.md) instead of, or
+// in addition to, inline metadata.import blocks on resource specs.
+//
+// Implementations must be nil-safe: LoadImportManifest(nil) returns nil.
+type ImportManifestConsumer interface {
```
abhimanyubabbar:
```suggestion
type ImportMetadataLoader interface {
     LoadImportMetadata(m *specs.WorkspaceImportMetadata) error
}

```

abhimanyubabbar: I was thinking the `ImportManifest` contains the ImportMetadata and therefore the interface could represent the ImportMetadataLoader as the name. Wdyt ?

cli/internal/project/importmanifest/provider.go
```diff
@@ -106,3 +106,31 @@ func (p *Provider) RuleDocEntries() []docs.RuleDocEntry {
...
+// ImportManifest returns the aggregated workspace entries, scoped to the
+// active workspace, wrapped in the shared WorkspacesImportMetadata type.
+//
+// Returns nil when no manifest specs were loaded. An empty activeWorkspaceID
+// means "all workspaces" (unscoped — used by validate); otherwise only the
+// entries for the active workspace are returned (D13: at-most-one remote per
+// URN reaches a handler). Returns nil when no entry matches the active
+// workspace. The orchestrator broadcasts this into the resource provider tree
+// via ImportManifestConsumer.
+func (p *Provider) ImportManifest(activeWorkspaceID string) *specs.WorkspacesImportMetadata {
```
abhimanyubabbar: To keep the signature clean, can we not have simply all the import manifest being returned and the caller then reducing it to per workspace ?

To me the element of suprise of passing in a workspace which is empty and then returning all is magic and shouldn't be done here. For now we can think about returning everything and let the caller do the filteration on it ?

cli/internal/providers/retl/provider.go
```diff
@@ -50,6 +50,22 @@ func New(client retlClient.RETLStore) *Provider {
...
 }
 
+// LoadImportManifest fans the aggregated manifest out to every registered
+// handler so URN → remote-ID mappings from a central import-manifest file
+// populate the same import-metadata map that inline metadata.import does.
+// Nil-safe.
+func (p *Provider) LoadImportManifest(m *specs.WorkspacesImportMetadata) error {
+	if m == nil {
+		return nil
+	}
```
abhimanyubabbar: Maybe by returning an actual object with empty workspaces would help us to remove this check from everywhere right ?

cli/internal/provider/provider.go
abhimanyubabbar: Are we planning to add it to the overall interface of the Providers later in subsequent PR ?

Given we have already handled the implementation for it why not make it mandatory for the Provider to declare it and then add it to the ProjectProvider interface ?

abhimanyubabbar (commented):
Some comments and clarifications

cli/internal/provider/provider.go
Jayachand: Prefer ImportManifestLoader. This calls LoadImportManifest -> calls underlying handler LoadImportMetadata

Jayachand: Made it mandatory

cli/internal/providers/retl/provider.go
Jayachand: removed

cli/internal/project/importmanifest/provider.go
Jayachand: moved logic handling to handleValidation

abhimanyubabbar (approved):

PR #634, opened 2026-06-25
feat(import): emit import-manifest entries from export and write import-manifest yaml

Jayachand:
## 🔗 Ticket

Part of [DEX-365](https://linear.app/rudderstack/issue/DEX-365/produce-manifest-file-on-import-command)

> **Scope note:** this PR delivers the export-entry plumbing + manifest file write + header comment. It **deliberately keeps inline `metadata.import` in emitted specs** for now; the "clean specs" (remove inline import) and the import→apply E2E round-trip from DEX-365 are deferred to a follow-up. The ticket has been updated to reflect this split.

---

## Summary

Changes the import/export write path so every provider returns the URN → remote-ID mappings it already embeds inline **also** as a flat `[]importmanifest.ImportEntry`, threaded up to the importer. The importer aggregates them via a new `importmanifest.BuildSpec`/`BuildNode` and writes a single project-level `imported/import-manifest.yaml`, prefixed with a header comment that discourages hand-edits. Inline `metadata.import` emission is unchanged, so behaviour is additive: the manifest is written *in addition to* today's output.

---

## Changes

* **`Exporter.FormatForExport` now returns `([]writer.FormattableEntity, []importmanifest.ImportEntry, error)`** — threaded through `Exporter`, the provider `Handler` interface, `HandlerImpl`, retl `resourceHandler`, event-stream `handler`, datacatalog `entityProvider`, and every aggregator (`BaseProvider`, `CompositeProvider`, `BaseHandler`, the three sub-provider aggregators).
* **Leaf implementers build entries** from the same tuples used for inline metadata: export strategies (single/multi spec), retl/sqlmodel, event-stream/source, datagraph, transformations (library + transformation), and all 5 datacatalog `importremote` providers. Small shared helpers (`importEntriesFromWorkspace`, `handlers.ImportEntriesFromWorkspace`) avoid duplication.
* **New `importmanifest.ImportEntry` + `BuildSpec(entries)`** — groups entries by workspace (first-seen order) into an `import-manifest` spec that round-trips through the existing `parseWorkspaces`.
* **New `importmanifest.BuildNode(entries)` + `ManifestHeaderComment` + `FileName`** — produces a `*yaml.Node` carrying the header comment.
* **Formatter `*yaml.Node` passthrough** — `YAMLFormatter.Format` uses a `*yaml.Node` as-is instead of re-encoding, so `HeadComment` survives (re-encoding would drop it). Non-node input is unchanged (regression-tested).
* **Importer** writes `imported/import-manifest.yaml` alongside the resource specs.
* Test mock + all affected handler/strategy tests updated to the 3-value return; entry assertions added to the export-strategy tests.

Generated file starts with:
```yaml
#  This is a project-level import manifest, generated by rudder-cli import.
#  It maps local resource URNs to remote IDs so apply reconciles them as imports.
#  Do not edit unless required.
version: "rudder/v1"
kind: "import-manifest"
...
```

---

## Testing

* **Unit**: `BuildSpec` grouping + `parseWorkspaces` round-trip; `BuildNode` header rendered through the formatter; formatter node-passthrough regression; export-strategy entry assertions; all existing handler/provider export tests updated and passing.
* `make test` — no failures. `make lint` — 0 issues. `go build ./...` clean.
* **E2E deferred**: `cli/tests/` runs against a live API; the import→apply round-trip is part of the deferred DEX-365 remainder.

---

## Risk / Impact

Low. Additive: the manifest file is written *in addition to* existing output and inline `metadata.import` is preserved, so current `import workspace` behaviour is unchanged apart from the extra file. The `FormatForExport` signature change is internal (no exported CLI surface).

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


abhimanyubabbar (approved):
The changes generally look good to me. Nothing specific stands out.
:shipit:

PR #635, opened 2026-06-25
feat(cli): add destination definition registry framework

abhimanyubabbar:
## 🔗 Ticket

Resolves [RUD-2866](https://linear.app/rudderstack/issue/RUD-2866)

---

## Summary

Adds the core destination definition registry framework for the CLI destination provider. This introduces versioned destination definitions with JSON Schema validation, local/API config conversion, and connection-mode metadata extraction — without wiring specific destination types (WEBHOOK, GA4, S3) yet.

---

## Changes

- Add `definitions` package with `DestinationDefinition`, `RegisteredDefinition`, and `Registry` keyed by `(Type, Version int64)`
- Add JSON Schema validator (Draft-07 via `kaptinlin/jsonschema`) with path-aware `ConfigError` collection
- Add config property DSL and `LocalToAPI` / `APIToLocal` converter (adapted from terraform-provider-rudderstack)
- Add shared consent management schema helpers in `common/`
- Add conversion test harness in `testutil/`
- Bump Go to 1.26.4 and add `kaptinlin/jsonschema`, `tidwall/gjson` dependencies

---

## Testing

- `make lint` — pass
- `go test ./cli/internal/providers/destination/definitions/...` — pass
- Unit tests cover registry API, schema compilation, validation (path + message), definition registration, converter round-trips, and consent management helpers

---

## Risk / Impact

**Low** — new isolated package under `cli/internal/providers/destination/`; no apply-cycle or CLI command changes in this PR.

Notes: Per-destination definitions and provider wiring are follow-up work.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


dhawal1248 (commented):
Need a call to go over the PR @abhimanyubabbar  and design decisions we took

cli/internal/providers/destination/definitions/validator.go
```diff
@@ -0,0 +1,81 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/go-viper/mapstructure/v2"
+
+	prulefuncs "github.com/rudderlabs/rudder-iac/cli/internal/provider/rules/funcs"
+	"github.com/rudderlabs/rudder-iac/cli/internal/validation/rules"
+)
+
+func validateConfigModel(config map[string]any, configType reflect.Type, basePath string) []ConfigError {
```
sivashanmukh, line 14: Was this an intentional re-design? Or have we deferred using `kaptinlin/jsonschema` speced out in https://linear.app/rudderstack/issue/RUD-2866/implement-core-destination-definition-registry-framework#step-4-schema-validator-7ed50804 for later?

cli/internal/providers/destination/definitions/common/consent_management.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	merged, err := json.Marshal(baseSchema)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling merged schema: %w", err)
+	}
+	return json.RawMessage(merged), nil
+}
+
+func camelToSnake(s string) string {
```
sivashanmukh, line 111: Claude found that this is a naive convertor that would convert `useNativeSDK` → `use_native_s_d_k`

Given that its only used for sourceType conversion today where none of the types have this kind of naming, but it can be potentially something to just comment over the function for any future users.

cli/internal/providers/destination/definitions/converter/configproperty.go
```diff
@@ -0,0 +1,336 @@
...
+	return ConfigProperty{
+		FromLocalFunc: discriminatorValue(apiKey, values),
+		ToLocalFunc:   func(local, config string) (string, error) { return local, nil },
+	}
+}
+
+// DiscriminatorValues maps local config keys to API discriminator values.
+type DiscriminatorValues map[string]any
+
+func ArrayWithStrings(rootAPIKey, nestedAPIField, localKey string) ConfigProperty {
```
sivashanmukh, line 86: Are you planning to use this anywhere?

cli/internal/providers/destination/definitions/registry.go
```diff
@@ -0,0 +1,87 @@
+package definitions
+
+import (
+	"fmt"
+	"slices"
+)
+
+type definitionKey struct {
+	Type    string
+	Version int64
```
sivashanmukh, line 10: In the spec you have defined using string here: https://linear.app/rudderstack/issue/RUD-2866/implement-core-destination-definition-registry-framework#internal-types-42ec60a0

Is there some other reason for using int64 here instead?

cli/internal/providers/destination/definitions/validator.go
```diff
@@ -0,0 +1,81 @@
...
+
+	errors := findUnknownKeys(config, configType, basePath)
+
+	decoded := reflect.New(configType).Interface()
+	if err := decodeConfig(config, decoded); err != nil {
+		errors = append(errors, ConfigError{
+			Path:    basePath,
+			Message: err.Error(),
+		})
+		return errors
```
sivashanmukh, line 31: `structValidationErrors` gets skipped from final return here if `decodeConfig` emits any errors.

sivashanmukh (changes requested):

cli/internal/providers/destination/definitions/validator.go
abhimanyubabbar: This was discovered during my refinement of spec and also discussion with Fotis earlier. I initially was of a view point that we can get by with using the jsonschema with minor changes but the code to parse the errors and then refactor them as per our internal error messaging was too huge.

This made me rethink the strategy and align with discussion with Fotis of how we can already leverage the struct validation we have. This seemed easier to onboard with minimal changes.

cli/internal/providers/destination/definitions/converter/configproperty.go
abhimanyubabbar: Not at this moment, it's verbatim copy-pasted from terraform module and hopefully when onboarding actual destinations we get to use it.

cli/internal/providers/destination/definitions/registry.go
abhimanyubabbar: Yes, when we started out initially Version was a string field "v1", "v2" etc. During the course of developing the changes with Sai and dhawal on the destination definition, they corrected it to int "1", "2" and hence the deviation in my spec and implementation.

cli/internal/providers/destination/definitions/validator.go
abhimanyubabbar: I'll check it out, thanks

cli/internal/providers/destination/definitions/common/consent_management.go
abhimanyubabbar: This conversion to me is wrong. I'll check and update the logic
Thanks for pointing this out.

cli/internal/providers/destination/definitions/validator.go
sivashanmukh: Makes sense. In that case, whenever a new definition version is supported in cli, we need some way to ensure that the json-schemas are translated accurately into the struct validations.

cli/internal/providers/destination/definitions/converter/configproperty.go
sivashanmukh: The reason I asked about this is because of what claude flagged in its review:

`ArrayWithStrings skip on len(contents)==0, and SkipZeroValue drops explicit "". Inherited TF behavior, fine for apply, but for the differ (RUD-2864) an explicitly-set debug_mode: false or empty whitelist will vanish on round-trip and can cause perpetual diffs. Note it for the handler ticket.`

In retrospect, I believe this is true for `ArrayWithObjects` as well for false/[] values. Worth dropping a note for when differ piece comes in and capture it in tests to ensure avoiding any unneccessary perpetual diffs.

cli/internal/providers/destination/definitions/validator.go
abhimanyubabbar: This would be done via skills which we will write and then the tests

cli/internal/providers/destination/definitions/converter/configproperty.go
abhimanyubabbar: I'll check it out.

cli/internal/providers/destination/definitions/common/consent_management.go
abhimanyubabbar: @sivashanmukh the only place this function is used is within the consent management to convert the sourceType. With this in mind it should be okay to agree that the issue with useNativeSDK won't happen. Wdyt ?

cli/internal/providers/destination/definitions/converter/configproperty.go
abhimanyubabbar: Is it okay when we start onboarding the actual destinations, we then check it out. I'll keep a note of it

cli/internal/providers/destination/definitions/validator.go
abhimanyubabbar: Given structValidationErrors work on the decoded config which itself is hydrated on the decodeConfig function, it should be correct right or did I miss the point completely ?

cli/internal/providers/destination/definitions/common/consent_management.go
sivashanmukh: Makes sense. Could you please just add a note as a comment with this as an example for any future users of this to be aware of the effect?

sivashanmukh (approved):

PR #641, opened 2026-07-01
fix(typer): TS interface field keys verbatim — Option 2 (quoted serial-name fields)

sivashanmukh:
## 🔗 Ticket

Relates to [DAW-3732](https://linear.app/rudderstack/issue/DAW-3732)

---

## Summary

RudderTyper v2's TypeScript generator camelCased tracking-plan property keys (`product_id` → `productId`) when emitting interface field names, but the generated code passes the `props` object verbatim to `analytics.track()`. The camelCased key therefore went on the wire, violating the tracking plan. Swift/Kotlin avoid this by keeping the original key. This addresses `DAW-3732`.

This is **Option 2 of 2 (draft, for comparison): quoted serial-name fields — field name == wire key, no runtime remap.** The interface field name is set to the plan's original property key so the emitted field name matches the key sent on the wire; no runtime remapping of the `props` object is required. The key is quoted only when it is not a valid TS identifier.

---

## Changes

- `getOrRegisterInterfacePropertyName` (`cli/internal/typer/generator/platforms/typescript/generator.go`) now emits the plan property key verbatim as the interface field name, quoting only when it is not a valid TS identifier. Dropped the `NameRegistry` collision registration for this path since JSON object keys are unique within a schema.
- `FormatPropertyName` is unchanged — it is still used for method names; only interface FIELD keys change.
- Regenerated the golden file `cli/internal/typer/generator/platforms/typescript/testdata/RudderTyper.ts`; snake_case plan properties (e.g. `product_id`, `page_type`, `first_name`) now appear verbatim as interface fields.
- Updated unit-test expectations in `track_test.go` and `variants_test.go` that asserted camelCased discriminator field names (`pageType` → `page_type`, `deviceType` → `device_type`).

---

## Testing

- Unit tests: `go test ./internal/typer/...` — all pass.
- `make lint` — 0 issues.
- Manual: verified the regenerated `RudderTyper.ts` now emits `product_id`/`page_type`/`first_name` interface fields matching the wire keys.

---

## Risk / Impact

Medium

Notes: This changes the generated interface field identifiers from camelCase to the plan's original property keys (a shape change for existing v2 TypeScript users who reference the generated field names). It fixes wire correctness — the key emitted in the generated interface now matches what is sent to `analytics.track()`. Method names are unaffected.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated
- [ ] No breaking changes (or documented) — documented above: interface field identifiers change shape for existing v2 TS users

### DIFF
--- a/cli/internal/typer/generator/platforms/typescript/generator.go
+++ b/cli/internal/typer/generator/platforms/typescript/generator.go
@@ -175,23 +175,17 @@ func getOrRegisterEventMethodName(rule *plan.EventRule, nr *core.NameRegistry) (
 // inside an interface, plus a `quoted` flag indicating whether the template
 // must wrap it in double quotes.
 //
-// When the JSON key sanitises to a valid TS identifier, it is registered with
-// the NameRegistry under a per-interface scope so duplicates after sanitisation
-// trigger the collision handler. When it does not (e.g. "用户名", "user-id",
-// "first name"), the original key is preserved verbatim and emitted as a
-// quoted property; registration is unnecessary because JSON object keys are
-// already unique within a single schema.
+// Emit the plan's property key verbatim so the generated field name matches the
+// key sent on the wire; the props object is passed to analytics.track()
+// unchanged, so camelCasing the field name would ship a key that violates the
+// tracking plan. Quote only when the key is not a valid TS identifier
+// (e.g. "用户名", "user-id", "first name"). JSON object keys are unique within a
+// single schema, so no NameRegistry collision registration is needed here.
 func getOrRegisterInterfacePropertyName(interfaceName, propName string, nr *core.NameRegistry) (name string, quoted bool, err error) {
-	formatted := FormatPropertyName(propName)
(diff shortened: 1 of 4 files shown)


sivashanmukh:
Closing in favour of #642 (Option 1: `SerialName` + runtime `applyKeyMap`).

Rationale, for the record: RudderTyper v2 landed with camelCase identifiers on all four platforms, and Swift/Kotlin pair that with the original plan key carried through to serialization (`SerialName` + `toProperties()`, `put("first_name", ...)`). The TypeScript generator implemented the naming half and not the serialization half — so this is an incomplete implementation of the v2 decision, not a reason to change the decision.

Option 2 (this PR) would have been correct on the wire, but it changes the generated interface field identifiers, which breaks v2 TypeScript users at compile time on top of the wire change that both options carry. Breaking the same users twice in one release cycle isn't worth the smaller diff.

Keeping this branch around as the documented alternative rather than deleting it.

PR #642, opened 2026-07-01
feat(typer): TS key remap via SerialName + runtime applyKeyMap (DAW-3732) [Option 1 of 2]

sivashanmukh:
## 🔗 Ticket

Relates to [DAW-3732](https://linear.app/rudderstack/issue/DAW-3732)

---

## Summary

DAW-3732: RudderTyper v2's TypeScript generator camelCases tracking-plan property keys (`product_id` → `productId`) and then passes the `props` object **verbatim** to `analytics.track(...)`. The camelCased key therefore lands on the wire, violating the tracking plan. Swift and Kotlin already avoid this by carrying the ORIGINAL key alongside the ergonomic identifier and rebuilding the payload before the SDK call (Swift: `SerialName` + `toProperties()`; Kotlin: `put("first_name", firstName)`).

This PR is **Option 1 of 2 (draft, for comparison): SerialName + runtime remap, mirroring Swift/Kotlin.** It keeps the camelCase interface field names for ergonomics but remaps keys back to the original plan names at runtime, just before the SDK call.

---

## Changes

- **`SerialName` field** added to `TSInterfaceProperty` holding the original plan key. `Name` stays camelCase (no camelCasing reverted). Populated wherever properties are built: event interfaces (`buildInterfaceWithNested`), custom-type interfaces (`buildInterfaceFromSchema`), and variant case interfaces (`variants.go`).
- **Per-interface key maps** (`keymap.go`): `collectKeyMaps` emits a `const <Interface>KeyMap: KeyMap = { camelName: "serial_name", ... }` for every interface that has ≥1 renamed key **or** a nested object that itself needs remapping. Interfaces whose keys already match the wire produce no map.
- **Shared runtime helper** `applyKeyMap(obj, map)` emitted once per generated file, plus `applyKeyMapNested`. It rewrites keys recursively (objects and arrays of objects), mirroring Swift's nested `toProperties()`.
- **Track call site rewrite**: `props as unknown as SDKApiObject` → `applyKeyMap(props, <Interface>KeyMap) as unknown as SDKApiObject` for track methods whose typed interface needs a map. Methods with no remap keep passing `props` directly.
- **Topological ordering** of key-map constants so a nested reference is initialized before its parent (TS module-load `const` TDZ safety).
- Golden file `testdata/RudderTyper.ts` regenerated via `make typer-typescript-update-testdata`.

---

## Testing

- Unit tests: `go test ./internal/typer/...` passes.
- New `keymap_test.go` covers: only-renamed interfaces get a map, nested recursion + dependency ordering, type-decoration stripping (`Foo[]`, `Array<Foo>`), and the track-arg rewrite (remapped vs verbatim).
- Updated struct-equality expectations in `track_test.go`, `variants_test.go`, `typemapping_test.go` to include the new `SerialName` field (tests were not weakened — the exact original keys are asserted).
- `make lint`: 0 issues. `go build ./...` clean.
- Verified in the regenerated golden: interface fields stay camelCase (e.g. `productId`, `ipAddress`, `deviceType`); the deepest recursion chain `UserSignedUpKeyMap → UserSignedUpContextKeyMap → UserSignedUpContextNestedContextKeyMap → CustomTypeUserProfileKeyMap` is emitted bottom-up and applied at the `trackUserSignedUp` call site.

---

## Risk / Impact

Low / Medium.

- **Added runtime surface**: every generated file that has ≥1 remapped event now carries an `applyKeyMap`/`applyKeyMapNested` helper, a `KeyMap`/`KeyMapValue` type pair, and one `const …KeyMap` per interface needing remap. This is per-generated-file code the SDK user ships. It is dead-code-eliminable only if unreferenced; some variant-case maps are emitted but unused when their union isn't remapped (see below).
- **Nested-object handling status — full-recursive for interface references, shallow+documented for unions**: nested objects referenced by a single interface name (inline event nested objects, object custom types) ARE remapped recursively via nested key maps. **Discriminated-union props are NOT remapped**: a track method whose props type is a union alias (e.g. `EventWithVariants = A | B | Default`) still forwards `props` verbatim, because correct union remapping needs per-branch dispatch on the discriminator, which this draft intentionally does not implement. Likewise, `identify`/`group`/`page` traits/properties are left verbatim in this draft (the wiring currently targets track — the canonical case). These are the honest limitations of Option 1 as drafted.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/typer/generator/platforms/typescript/keymap.go
+++ b/cli/internal/typer/generator/platforms/typescript/keymap.go
@@ -1,0 +1,219 @@
+package typescript
+
+import (
+	"sort"
+	"strings"
+)
+
+// keyMapSuffix is appended to an interface name to form its generated key-map
+// constant, e.g. "UserSignedUp" → "UserSignedUpKeyMap".
+const keyMapSuffix = "KeyMap"
+
+// collectKeyMaps walks every generated interface and emits a per-interface
+// camelCase→serial-name key map (TSKeyMap) for those that actually need one.
+//
+// An interface needs a map when at least one of its properties either:
+//   - renames its wire key (camelCase Name differs from SerialName), or
+//   - references another interface that itself needs a map (so the nested
+//     object's keys get remapped recursively — mirroring Swift's nested
+//     toProperties()).
+//
+// Maps are attached to ctx.KeyMaps (sorted by name for deterministic output)
+// and the set of interface names that have a map is returned so callers can
+// decide whether a method's props need remapping. This is the single source of
+// truth for "does interface X need remapping".
+func collectKeyMaps(ctx *TSContext) map[string]bool {
+	interfaces := allInterfaces(ctx)
+
+	// byName lets property-type resolution find a referenced interface. Names of
+	// interfaces that need a map is computed to a fixed point below, since a
+	// parent needs a map if a child does.
+	byName := make(map[string]*TSInterface, len(interfaces))
+	for i := range interfaces {
+		byName[interfaces[i].Name] = interfaces[i]
+	}
+
+	needsMap := make(map[string]bool)
+	// Fixed-point iteration: repeat until no interface flips to "needs map".
+	// Bounded by the number of interfaces (each pass can only add flags).
+	for {
+		changed := false
+		for name, iface := range byName {
+			if needsMap[name] {
+				continue
+			}
+			if interfaceNeedsMap(iface, needsMap) {
+				needsMap[name] = true
+				changed = true
+			}
+		}
+		if !changed {
+			break
+		}
+	}
+
+	maps := make([]TSKeyMap, 0, len(needsMap))
+	for _, iface := range interfaces {
+		if !needsMap[iface.Name] {
+			continue
+		}
+		maps = append(maps, buildKeyMap(iface, needsMap))
+	}
+
+	ctx.KeyMaps = orderKeyMaps(maps)
+	return needsMap
+}
+
+// orderKeyMaps returns the maps ordered so that every map appears after the
+// maps it references. TS `const` initializers run at module load, so a nested
+// reference to a map defined later would hit the temporal dead zone — emitting
+// dependencies first avoids that. Within a dependency tier the order is
+// alphabetical for deterministic output.
+func orderKeyMaps(maps []TSKeyMap) []TSKeyMap {
+	byName := make(map[string]TSKeyMap, len(maps))
+	names := make([]string, 0, len(maps))
+	for _, m := range maps {
+		byName[m.Name] = m
+		names = append(names, m.Name)
+	}
+	sort.Strings(names)
+
+	var (
+		ordered  []TSKeyMap
+		emitted  = make(map[string]bool)
+		visiting = make(map[string]bool)
+		visit    func(name string)
+	)
+	visit = func(name string) {
+		if emitted[name] || visiting[name] {
+			return // visiting guard also breaks any (unexpected) cycle
+		}
+		m, ok := byName[name]
+		if !ok {
+			return
+		}
+		visiting[name] = true
+		for _, e := range m.Entries {
+			if e.NestedMapName != "" {
+				visit(e.NestedMapName)
+			}
+		}
+		visiting[name] = false
+		emitted[name] = true
+		ordered = append(ordered, m)
+	}
+	for _, name := range names {
+		visit(name)
+	}
+	return ordered
+}
+
+// wireTrackKeyMaps points each track method at the key map for its typed props
+// interface, rewriting the props SDK argument to route through applyKeyMap so
+// the wire payload uses original plan keys. Methods whose interface needs no
+// remap (or whose props are an open Record, not a typed interface) are left
+// forwarding props verbatim.
+func wireTrackKeyMaps(ctx *TSContext, needsMap map[string]bool) {
+	for mi := range ctx.AnalyticsMethods {
+		m := &ctx.AnalyticsMethods[mi]
+		if m.SDKMethodName != "track" || len(m.MethodArguments) == 0 {
+			continue
+		}
+		ifaceName := m.MethodArguments[0].Type
+		if !needsMap[ifaceName] {
+			continue
+		}
+		mapName := ifaceName + keyMapSuffix
+		m.PropsKeyMapName = mapName
+		ctx.UsesApplyKeyMap = true
+		// The props SDK argument is the one referencing `props` — rewrite it to
+		// remap keys first. Other args (event name, options, callback) are
+		// untouched.
+		for ai := range m.SDKArguments {
+			if strings.HasPrefix(m.SDKArguments[ai].Value, "props ") {
+				m.SDKArguments[ai].Value = "applyKeyMap(props, " + mapName + ") as unknown as " + sdkApiObjectAlias
+			}
+		}
+	}
+}
+
+// allInterfaces returns pointers to every interface the generator produced,
+// across event, nested, custom, and variant-case buckets.
+func allInterfaces(ctx *TSContext) []*TSInterface {
+	var out []*TSInterface
+	for i := range ctx.CustomInterfaces {
+		out = append(out, &ctx.CustomInterfaces[i])
+	}
+	for i := range ctx.NestedInterfaces {
+		out = append(out, &ctx.NestedInterfaces[i])
+	}
+	for i := range ctx.Interfaces {
+		out = append(out, &ctx.Interfaces[i])
+	}
+	for gi := range ctx.VariantTypes {
+		for ci := range ctx.VariantTypes[gi].CaseInterfaces {
+			out = append(out, &ctx.VariantTypes[gi].CaseInterfaces[ci])
+		}
+	}
+	return out
+}
+
+// interfaceNeedsMap reports whether iface has any property that must be
+// remapped, given the current known set of interfaces that need a map.
+func interfaceNeedsMap(iface *TSInterface, needsMap map[string]bool) bool {
+	for _, p := range iface.Properties {
+		if p.Name != p.SerialName {
+			return true
+		}
+		if referencedInterfaceName(p.Type, needsMap) != "" {
+			return true
+		}
+	}
+	return false
+}
+
+// buildKeyMap constructs the TSKeyMap for an interface known to need one. Only
+// properties that rename or reference a nested map produce entries; identity
+// keys are omitted so applyKeyMap leaves them untouched.
+func buildKeyMap(iface *TSInterface, needsMap map[string]bool) TSKeyMap {
+	km := TSKeyMap{Name: iface.Name + keyMapSuffix}
+	for _, p := range iface.Properties {
+		nested := referencedInterfaceName(p.Type, needsMap)
+		if p.Name == p.SerialName && nested == "" {
+			continue
+		}
+		entry := TSKeyMapEntry{FieldName: p.Name, SerialName: p.SerialName}
+		if nested != "" {
+			entry.NestedMapName = nested + keyMapSuffix
+		}
+		km.Entries = append(km.Entries, entry)
+	}
+	return km
+}
+
+// referencedInterfaceName returns the name of a single interface referenced by
+// a TS type expression when that interface needs a map, else "". It strips
+// array/optional decorations (`Foo[]`, `Array<Foo>`, `Foo | null`) so a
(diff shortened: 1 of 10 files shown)


cli/internal/typer/generator/platforms/typescript/keymap.go
```diff
@@ -0,0 +1,219 @@
...
+	t = strings.TrimSuffix(t, "[]")
+	if strings.HasPrefix(t, "Array<") && strings.HasSuffix(t, ">") {
+		t = strings.TrimSuffix(strings.TrimPrefix(t, "Array<"), ">")
+	}
+	t = strings.TrimSpace(t)
+	// Bail on unions/generics — only a bare interface reference is safe to remap.
+	if strings.ContainsAny(t, "|<>") {
+		return ""
+	}
+	if needsMap[t] {
```
akashrpo, line 215: Arrays of objects don't get remapped. track({ profileList: [{ firstName: "x" }] }) → wire sends firstName, should be first_name.

cli/internal/typer/generator/platforms/typescript/keymap.go
```diff
@@ -0,0 +1,219 @@
...
+			continue
+		}
+		mapName := ifaceName + keyMapSuffix
+		m.PropsKeyMapName = mapName
+		ctx.UsesApplyKeyMap = true
+		// The props SDK argument is the one referencing `props` — rewrite it to
+		// remap keys first. Other args (event name, options, callback) are
+		// untouched.
+		for ai := range m.SDKArguments {
+			if strings.HasPrefix(m.SDKArguments[ai].Value, "props ") {
```
akashrpo, line 133: The props argument is located by `HasPrefix(Value, "props ")`. But that text is a hardcoded literal written at generator.go:944 (`"props as unknown as SDKApiObject"`), and nothing keeps the two in sync. If that line changes (rename param, tweak the cast, extra space), this match silently fails: `applyKeyMap` never gets added, keys go out camelCase again, and no test catches it.

Suggest a flag on the arg instead of string-matching:

```go
// context.go, TSSDKArgument
type TSSDKArgument struct {
    Value   string
    IsProps bool  // the properties/traits arg applyKeyMap wraps
}

// generator.go:944 and :955
TSSDKArgument{Value: "props as unknown as " + sdkApiObjectAlias, IsProps: true}

// keymap.go:133
if m.SDKArguments[ai].IsProps {
    m.SDKArguments[ai].Value = "applyKeyMap(props, " + mapName + ") ..."
}
```

Update the hand-built args in `TestWireTrackKeyMaps_RewritesPropsArg` to set `IsProps: true`.

cli/internal/typer/generator/platforms/typescript/testdata/RudderTyper.ts
```diff
@@ -415,6 +415,171 @@ export interface EventWithNameCamelCase1 {
...
+    pageData: "page_data",
+    pageType: "page_type",
+};
+
+const CustomTypeUserProfileKeyMap: KeyMap = {
+    firstName: "first_name",
+    lastName: "last_name",
+};
+
+const EventWithVariantsCaseDesktopKeyMap: KeyMap = {
```
akashrpo, line 507: EventWithVariantsCase: KeyMap generated but never referenced

cli/internal/typer/generator/platforms/typescript/keymap.go
```diff
@@ -0,0 +1,219 @@
...
+// See the nested-object limitation in the PR body.
+func referencedInterfaceName(tsType string, needsMap map[string]bool) string {
+	t := strings.TrimSpace(tsType)
+	t = strings.TrimSuffix(t, "[]")
+	if strings.HasPrefix(t, "Array<") && strings.HasSuffix(t, ">") {
+		t = strings.TrimSuffix(strings.TrimPrefix(t, "Array<"), ">")
+	}
+	t = strings.TrimSpace(t)
+	// Bail on unions/generics — only a bare interface reference is safe to remap.
+	if strings.ContainsAny(t, "|<>") {
```
akashrpo, line 212: Nested objects that can be "one of several shapes" (union types, like `featureConfig`) only get their outer key fixed. The keys inside them still go out camelCase.

Example: `track({ featureConfig: { featureFlag: true } })` sends `featureFlag` on the wire, but the plan expects `feature_flag`. Broken.

This code skips these because it can't tell which shape the object is at runtime, so it gives up. The maps to fix them are even generated but never used.

Not asking to fix it here (needs runtime shape-detection). Just flag it as a known correctness gap, and note it hits normal events too, not only union-typed track props like the PR says.

cli/internal/typer/generator/platforms/typescript/keymap.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return ordered
+}
+
+// wireTrackKeyMaps points each track method at the key map for its typed props
+// interface, rewriting the props SDK argument to route through applyKeyMap so
+// the wire payload uses original plan keys. Methods whose interface needs no
+// remap (or whose props are an open Record, not a typed interface) are left
+// forwarding props verbatim.
+func wireTrackKeyMaps(ctx *TSContext, needsMap map[string]bool) {
```
akashrpo, line 116: Do we scope for page/identify/group as well?

cli/internal/typer/generator/platforms/typescript/testdata/RudderTyper.ts
```diff
@@ -415,6 +415,171 @@ export interface EventWithNameCamelCase1 {
...
+};
+
+const EventWithVariantsDefaultKeyMap: KeyMap = {
+    deviceType: "device_type",
+    pageContext: "page_context",
+    profile: { key: "profile", map: CustomTypeUserProfileKeyMap },
+    untypedField: "untyped_field",
+};
+
+const PagePropertiesKeyMap: KeyMap = {
```
akashrpo, line 528: PagePropertiesKeyMap is generated but page() never applies it

cli/internal/typer/generator/platforms/typescript/testdata/RudderTyper.ts
```diff
@@ -415,6 +415,171 @@ export interface EventWithNameCamelCase1 {
...
+    profile: { key: "profile", map: CustomTypeUserProfileKeyMap },
+};
+
+const EventWithVariantsCaseMobileKeyMap: KeyMap = {
+    deviceType: "device_type",
+    pageContext: "page_context",
+    profile: { key: "profile", map: CustomTypeUserProfileKeyMap },
+};
+
+const EventWithVariantsDefaultKeyMap: KeyMap = {
```
akashrpo, line 521: This one as well

akashrpo (changes requested):

sivashanmukh:
@akashrpo heads up — I **force-pushed** this branch. It was 66 commits behind main and #681 had changed the generated constructor underneath it, so a rebase was needed before any of this could be verified. Your review comments were anchored to `112b41f`, which is no longer on the branch — the old commit is still reachable at `112b41f2a9190a1a0960440e7ab8dfabbec6794e` if you want to diff against what you reviewed. New head is `1eb82a5`.

Sorry for detaching the thread. Every point you raised is addressed below, so hopefully re-reviewing from scratch isn't too painful.

---

### Decision first

We went with this PR (Option 1) and **closed #641** (Option 2, verbatim field keys). Reasoning is on that PR: v2 landed with camelCase identifiers on all four platforms, and Swift/Kotlin pair that with the original key carried through to serialization. TypeScript implemented the naming half and not the serialization half — so this is an incomplete implementation of the v2 decision, not a reason to change the decision. Option 2 was correct on the wire but changed the generated interface field identifiers, breaking v2 TS users at compile time *on top of* the wire change both options carry.

### Your comments

1. **Arrays of objects aren't remapped** — fixed. The runtime helper already recursed into arrays; the real bug was that `profile_list` is typed as `CustomTypeProfileList`, an *alias* for `Profile[]`. Resolution now follows aliases and strips array/nullable decorations. Same fix picked up `addresses`, which we'd both missed.
2. **Union-typed nested objects only get the outer key fixed** — fixed, and you were right that it hits normal events too, not just union-typed track props as the PR body claimed. It turned out not to need runtime shape-detection: the camelCase identifier is a pure function of the plan key, so sibling branches can never map the same field to different keys, and the branch maps merge into one map named after the union alias. `collectKeyMaps` errors rather than guessing if a merge ever does conflict.
3. **`HasPrefix(Value, "props ")` is a silent-failure path** — taken, slightly generalized. `propsArg(valueFmt, expr)` records the expression plus its surrounding format, because `group` nests its traits inside `withRudderTyperContext(...)` and a bare `IsProps bool` couldn't express that. The string match is gone.
4. **`EventWithVariantsCaseKeyMap` / `PagePropertiesKeyMap` generated but never referenced** — gone at the root. Variant case interfaces no longer get standalone maps (unreachable once the alias map exists), and `PagePropertiesKeyMap` is now actually applied.
5. **"Do we scope for page/identify/group as well?"** — yes, all four now go through one uniform path (`PropsTypeName` on the method + a rewrite pass over `DispatcherBranches`). `identify`/`group` emit no map for the reference plan because its traits genuinely have no renames; the wiring is covered by unit test instead.

### Verification

The gap here was that nothing measured the wire payload, which is how four shapes shipped wrong. Added `wireEquivalence.test.ts` to the validator: it drives track / variants / page / identify / group and fails on any wire key not in the plan.

| | before | after |
|---|---|---|
| `wireEquivalence.test.ts` | 2 / 5 | **5 / 5** |
| full validator suite | 3 files failing | **37 / 37**, typecheck clean |
| `make typer-typescript-validate` | fail | **pass** |

Validator expectations across track/variants/multitype/page were asserting the old camelCase payload — rewritten from the generator's own key maps. Call sites stay camelCase; that's the ergonomics this option preserves.

Generated file goes 769 → 899 lines: up 130, but 35 *fewer* than the previous draft despite covering four more shapes, since the dead per-case maps are gone.

`make lint` clean. Note `cli/tests` E2E fails identically on unmodified main locally (needs live credentials) — unrelated.

One thing worth your eye specifically: the merge-safety argument in point 2. If you think two branches *can* disagree in a way I've ruled out, the guard should become the primary mechanism rather than the fallback.

akashrpo:
**Blocking**

1. **Variant map collision** (`keymap.go:115`): `buildKeyMap` skips identity fields, so `first_name` in one branch and `firstName` in another can merge into `firstName -> first_name`, silently corrupting the second branch. Reproduced by running `collectKeyMaps`. Identity entries should be included during merge so this collision is detected/preserved.

2. **`wireEquivalence.test.ts` doesn't exercise the fix in CI**: `CAMEL_INPUT` requires `EQUIV_CAMEL=1`, but nothing in the test target sets it. Default tests pass on `main` too. Also, identify/group traits don't contain any renamable fields. Either remove this test or run both paths with actual renamed traits.

3. **Multi-type object properties fail generation** (`keymap.go:322`): `A | B` and `Array<A | B>` hit the inline-union guard and error. Reproduced by running `collectKeyMaps`. The reference plan doesn't hit this, but multi-object-type properties are valid plan input. Branch maps should be merged here instead of rejecting the type.

**Should fix**

4. **Recursive types generate invalid self-referencing consts** (`keymap.go:166`): a customer plan with `Node -> Node` generates a `NodeKeyMap` referencing itself during initialization, causing a TDZ `ReferenceError` on import. Either make nested maps lazy or error on cycles.

5. **Unused key maps are emitted** (`keymap.go:41`): an unreferenced object custom type can still generate a key map and break consumers with `noUnusedLocals`. Only emit maps reachable from wired methods.

6. **Wire behavior change has no migration/process coverage**: this changes what v2 TypeScript customers send on the wire, but the PR doesn't include a CHANGELOG/docs update or an opt-out/migration path. This should be addressed explicitly before merging.

**Minor**

* `applyKeyMap`: `out[key] = v; delete out[key]; out[entry] = v` reduces to `out[entry] = v`. The identity-map comment describes work that isn't needed.
* `applyKeyMapNested` is redundant. `applyKeyMap` already returns non-objects unchanged and handles arrays.
* Nested remapping can turn `Date`/`Map`/class instances into plain objects.
* `additionalProperties` can cause `first_name`/`firstName` key clobbering.
* `splitTopLevelUnion` (`keymap.go:434`) counts `<`, `(`, `[` inside quoted strings, so enum values containing these characters can skew depth tracking and cause resolution to bail early. Narrow, but can result in silent pass-through with remappable object unions.
* `propsArg` / `remapPropsArgs` format-string construction is fragile if interpolated type names contain `%`.

akashrpo (approved):

PR #655, opened 2026-07-03
feat(cli): add destination handler with CRUD lifecycle and state mapping

abhimanyubabbar:
## 🔗 Ticket

Resolves [RUD-2864](https://linear.app/rudderstack/issue/RUD-2864/destination-handler-models-crud-lifecycle-state-mapping-and-provider)

---

## Summary

Adds the destination provider and handler on `BaseHandler`, consuming the shipped `definitions.Registry` for snake_case ↔ camelCase config conversion at the API boundary, wiring CRUD lifecycle with transformation-link side effects, and registering the provider in the CLI app with an empty definition registry. Import/export and full validation rules remain stubbed for follow-up tickets.

---

## Changes

- Add `ExternalID` field to `client.Destination` for managed resource filtering (no `SetExternalID` method — deferred to RUD-2865)
- Create destination models (`DestinationSpec`, `DestinationResource`, `DestinationState`, `RemoteDestination` + `Metadata()`)
- Implement handler lifecycle: `ExtractResourcesFromSpec`, `Create`, `Update` (rejects immutable type change; reconciles transformation link), `Delete` (disconnects transformation first), `MapRemoteToState` (camelCase→snake_case via registry; graceful degradation when linked transformation is not CLI-managed), `LoadRemoteResources`/`LoadImportableResources` (ExternalID + registered-type filtering)
- Stub `Import` and `FormatForExport` (RUD-2865)
- Wire destination provider + empty registry in `dependencies.go` (production definitions deferred)
- Update project gatekeeper rule doc fragments (`duplicate-urn`, `metadata-syntax-valid`, `manifest-inline-conflict`) to cover `destination`/`rudder/v1` and mirror that in the ruledoc test scoping helpers

---

## Testing

- `make lint` — pass, 0 issues
- `make test` — pass, no failures (destination pkg coverage 77.9%)
- Unit tests for handler lifecycle (Create, Update incl. type-change rejection + transformation add/remove/replace/unchanged, Delete with/without link, MapRemoteToState incl. unregistered-type error + not-CLI-managed graceful path + empty-ExternalID skip, LoadRemote/LoadImportable filtering, Import/Export stubs)
- Model tests for mapstructure decode (incl. `definition_version` as int64), `#transformation:<id>` ref parsing, `RemoteDestination.Metadata()`
- Registry-integration test: full GA4 round-trip asserting config survives snake_case → camelCase → snake_case unchanged (the differ's correctness invariant), plus transformation link round-trip on both spec and state sides

---

## Risk / Impact

**Low** — New provider behind an empty registry; no production destination types registered yet, so the binary cannot apply a destination end-to-end until definitions are onboarded. E2E tests deferred. No changes to existing provider behavior.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/destination/handler.go
+++ b/cli/internal/providers/destination/handler.go
@@ -44,7 +44,7 @@ type HandlerImpl struct {
 // NewHandler builds a *DestinationHandler wired to the given client and registry.
 func NewHandler(c *client.Client, registry *definitions.Registry) *DestinationHandler {
 	h := &HandlerImpl{client: c, registry: registry}
-	return handler.NewHandler(h)
+	return handler.NewHandler[DestinationSpec, DestinationResource, DestinationState, RemoteDestination](h)
 }
 
 func (h *HandlerImpl) Metadata() handler.HandlerMetadata {
--- a/cli/internal/providers/destination/handler.go
+++ b/cli/internal/providers/destination/handler.go
@@ -84,7 +84,7 @@ func (h *HandlerImpl) ExtractResourcesFromSpec(_ string, spec *DestinationSpec) 
 func (h *HandlerImpl) Create(ctx context.Context, data *DestinationResource) (*DestinationState, error) {
 	apiConfig, err := h.localConfigToAPI(data.Type, data.DefinitionVersion, data.Config)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("converting local config to API: %w", err)
 	}
 
 	created, err := h.client.Destinations.Create(ctx, &client.Destination{
@@ -98,9 +98,14 @@ func (h *HandlerImpl) Create(ctx context.Context, data *DestinationResource) (*D
 		return nil, fmt.Errorf("creating destination: %w", err)
 	}
 
-	transformationID, err := h.connectTransformationIfPresent(ctx, created.ID, data.Transformation)
+	transformationID, err := h.syncTransformationLink(
+		ctx,
+		created.ID,
+		data.Transformation,
+		"", // no previous transformation ID
+	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("syncing transformation link: %w", err)
 	}
 
 	return &DestinationState{ID: created.ID, TransformationID: transformationID}, nil
@@ -192,7 +197,7 @@ func (h *HandlerImpl) MapRemoteToState(
 		return nil, nil, err
 	}
 
-	transformationRef, transformationID, err := h.mapTransformationRef(remote, urnResolver)
+	transformationRef, transformationID, err := h.transformationRef(remote, urnResolver)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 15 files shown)


Jayachand (approved):

cli/internal/providers/destination/handler.go
```diff
@@ -0,0 +1,444 @@
...
+		return nil, "", nil
+	}
+
+	transformationID := remote.Transformation.ID
+	urn, err := urnResolver.GetURNByID(ttypes.TransformationResourceType, transformationID)
+	if err != nil {
+		if err == resources.ErrRemoteResourceExternalIdNotFound {
+			// It might be a transformation which the upstream user's manually added but
+			// not managed by the CLI yet.
+			return nil, transformationID, nil
```
sivashanmukh: **Persisting the foreign transformation ID here can silently unlink a UI-managed transformation.**

On the not-CLI-managed path we drop the ref but still return `transformationID`, which `MapRemoteToState` stores as `state.TransformationID`. On a later apply where the user edits *any* spec field (config, `display_name`, `enabled`), the differ calls `Update`: `newData.Transformation` is `nil`, so `newTransformationID == ""`, but `oldState.TransformationID == "ui-trans"`, so `syncTransformationLink` hits the disconnect branch and unlinks the user's UI/API-added transformation as a side effect of an unrelated edit.

This diverges from the event-stream source handler cited as the precedent. That handler sets the link to `nil` and does **not** persist the foreign ID (`event-stream/source/handler.go:456-464`), precisely so it can't later be reconciled as if CLI-managed. The comment's "link re-applies on every run" also doesn't match the actual behavior (it disconnects). Suggest returning `""` for the state ID on this path to mirror event-stream.

cli/internal/providers/destination/handler.go
```diff
@@ -0,0 +1,444 @@
...
+
+// Create provisions the destination remotely, then links a transformation if the
+// spec-side PropertyRef was resolved by the apply framework.
+func (h *HandlerImpl) Create(ctx context.Context, data *DestinationResource) (*DestinationState, error) {
+	apiConfig, err := h.localConfigToAPI(data.Type, data.DefinitionVersion, data.Config)
+	if err != nil {
+		return nil, fmt.Errorf("converting local config to API: %w", err)
+	}
+
+	created, err := h.client.Destinations.Create(ctx, &client.Destination{
```
sivashanmukh, line 91: **`Create` omits `DefinitionVersion`, but `MapRemoteToState` reads `remote.Version` back as the definition version (L195/L211) and keys the registry lookup on it.**

If the API assigns a `Version` that differs from the spec's `definition_version` (e.g. it defaults to the latest registered definition), read-back yields `resource.DefinitionVersion != spec`, causing a perpetual diff, and `apiConfigToLocal` decodes with a different converter than `localConfigToAPI` encoded with. Every test uses `Version: 1` on both sides, so this path is untested. Can we confirm the API echoes back the same definition version the config was authored for, or should Create send the version explicitly?

cli/internal/providers/destination/handler.go
abhimanyubabbar: Very valid feedback

cli/internal/providers/destination/handler.go
abhimanyubabbar: Very Valid Feedback

PR #656, opened 2026-07-07
feat: add import support for the destination resources

abhimanyubabbar:
## 🔗 Ticket

Resolves [RUD-2865](https://linear.app/rudderstack/issue/RUD-2865)

---

## Summary

Implements `Import` and `FormatForExport` for the destination provider handler, completing the apply-cycle import/export lifecycle that was previously stubbed out.

---

## Changes

- `Import`: adopts an existing remote destination by fetching it (and its transformation link, since the single-resource `Get` endpoint doesn't embed it), reconciling config/transformation via the existing `Update` path, then setting the external ID last so a failed update never leaves a partially-adopted resource.
- `FormatForExport`: converts unmanaged remote destinations into importable YAML specs — converts config to local snake_case, masks registered secret keys with per-resource placeholders, and resolves a linked transformation to a `#transformation:<id>` reference (failing the export rather than silently falling back to a raw ID).
- `LoadImportableResources` now matches on the exact `(Type, Version)` pair instead of type alone, since only specific registered definition versions are convertible.
- Added `WorkspaceID` to `Destination` (API client) and `RemoteResourceMetadata` (model) so exported specs can be tied to an import-manifest workspace entry.
- `provider.go`: destinations own their full export path (`destinations/`), so the base import directory is left empty.

---

## Testing

- Unit tests added for `Import` (success, remote get error, update error skips setting external ID) and `FormatForExport` (empty collection, basic spec, secret masking, transformation resolution + resolution failure, unregistered type/version error).
- `go build ./...`, `go test ./cli/internal/providers/destination/... ./api/client/...`, and `make lint` all pass.

---

## Risk / Impact

Low. Scoped to the destination provider's import/export path; no changes to the apply/update/delete lifecycle.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/destination/handler.go
+++ b/cli/internal/providers/destination/handler.go
@@ -4,14 +4,17 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/rudderlabs/rudder-iac/api/client"
 	"github.com/rudderlabs/rudder-iac/cli/internal/namer"
 	"github.com/rudderlabs/rudder-iac/cli/internal/project/importmanifest"
+	"github.com/rudderlabs/rudder-iac/cli/internal/project/specs"
 	"github.com/rudderlabs/rudder-iac/cli/internal/project/writer"
 	"github.com/rudderlabs/rudder-iac/cli/internal/provider/handler"
 	"github.com/rudderlabs/rudder-iac/cli/internal/providers/destination/definitions"
+	"github.com/rudderlabs/rudder-iac/cli/internal/providers/transformations/handlers"
 	tmodel "github.com/rudderlabs/rudder-iac/cli/internal/providers/transformations/model"
 	ttypes "github.com/rudderlabs/rudder-iac/cli/internal/providers/transformations/types"
 	"github.com/rudderlabs/rudder-iac/cli/internal/resolver"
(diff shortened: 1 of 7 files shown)


cli/internal/providers/destination/handler.go
```diff
@@ -278,18 +285,150 @@ func (h *HandlerImpl) LoadImportableResources(ctx context.Context) ([]*RemoteDes
...
+	remote, err := h.client.Destinations.Get(ctx, remoteId)
+	if err != nil {
+		return nil, fmt.Errorf("getting destination during import: %w", err)
+	}
+
+	// The single-resource Get endpoint doesn't embed the transformation link
+	// (unlike the list endpoint used by LoadImportableResources), so it's
+	// fetched separately. No existing link is treated the same as an error
+	// here since the API has no "not found" sentinel for this sub-resource.
+	connectedTransformation, err := h.client.Destinations.GetTransformation(ctx, remoteId)
```
Jayachand, line 304: What happens in case of import destination resource which has no transformation connection? This api errors out when there is no connection.

Jayachand (approved):
Looks good other than connected transformation check

cli/internal/providers/destination/handler.go
abhimanyubabbar: This is handled, resolving it.

PR #661, opened 2026-07-08
feat: add support for the destination validations

abhimanyubabbar:
## 🔗 Ticket

Resolves [RUD-2853](https://linear.app/rudderstack/issue/RUD-2853/destination-validations)

---

## Summary

Adds syntactic and semantic validation rules for the `destination` resource kind, filling the rule hooks stubbed in #655. The syntactic rule validates the spec envelope (required fields, unknown keys), that the `(type, definition_version)` pair is registered, the transformation ref format, and the per-type config via the definition registry's struct-based `ValidateConfig`. The semantic rule validates that the referenced transformation exists in the resource graph and that `display_name` is unique across destinations. Scope refined from the original ticket: connection provider and connection-aware checks (source-type compatibility, dead config, completeness) are split to a follow-up ticket.

---

## Changes

- `destination/spec-syntax-valid` rule: envelope decode with all unknown envelope keys reported in one pass (`mapstructure.Metadata`), required-field struct validation, hard error on unregistered type/version with supported types/versions listed, `#transformation:<id>` format check, registry `ValidateConfig` mapping `ConfigError`s to `/spec/config/...` paths, and source-type key check (`connection_mode` / `use_native_sdk` / `consent_management` platform keys must be supported by the definition)
- `destination/semantic-valid` rule: transformation exists in resource graph, `display_name` uniqueness via `RawData()`
- `RegisteredDefinition.LocalSourceTypeKeys()` + exported `common.LocalSourceTypeKey` (camelCase → snake_case source type keys)
- `validate:"required"` tags on `DestinationSpec` (id, display_name, type, definition_version)
- Provider wiring for `SyntacticRules()` / `SemanticRules()` / `RuleDocEntries()` with embedded `*.docs.yaml` fragments for both rules
- Note: rules live in the `destination` package root rather than a `rules/` subpackage — the provider package hosts the spec model, so a subpackage would create an import cycle

---

## Testing

- Unit tests: full-slice `ValidationResult` assertions covering every syntactic step incl. early-return behavior, semantic graph cases (missing transformation, duplicate/unique display_name, foreign `RawData` skipped), provider wiring, and `LocalSourceTypeKeys`
- `make lint` (0 issues), `make test` (all packages pass), `make gen-rule-docs` (both rules join the generated catalog)
- Manual: built the CLI and ran `validate` against a sample destination spec — diagnostics land on exact YAML lines with correct rule IDs

---

## Risk / Impact

Low
Notes: validation-only change; no apply-cycle CRUD behavior touched. Until production definitions are registered, destination specs fail validation with "no destination types are currently supported" — intended gating per refinement decision.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/destination/model.go
+++ b/cli/internal/providers/destination/model.go
@@ -8,11 +8,11 @@ import (
 
 // DestinationSpec is the user-authored YAML representation of a destination.
 type DestinationSpec struct {
-	ID                string         `mapstructure:"id"`
-	DisplayName       string         `mapstructure:"display_name"`
-	Type              string         `mapstructure:"type"`
+	ID                string         `mapstructure:"id" validate:"required"`
+	DisplayName       string         `mapstructure:"display_name" validate:"required"`
+	Type              string         `mapstructure:"type" validate:"required"`
 	Enabled           bool           `mapstructure:"enabled"`
-	DefinitionVersion int64          `mapstructure:"definition_version"`
+	DefinitionVersion int64          `mapstructure:"definition_version" validate:"required"`
 	Transformation    string         `mapstructure:"transformation"` // scalar "#transformation:<id>" — object form deferred
 	Config            map[string]any `mapstructure:"config"`
 }
--- a/cli/internal/providers/destination/model.go
+++ b/cli/internal/providers/destination/model.go
@@ -9,7 +9,7 @@ import (
 // DestinationSpec is the user-authored YAML representation of a destination.
 type DestinationSpec struct {
 	ID                string         `mapstructure:"id" validate:"required"`
-	DisplayName       string         `mapstructure:"display_name" validate:"required"`
+	DisplayName       string         `mapstructure:"display_name" validate:"required,pattern=destination_display_name"`
 	Type              string         `mapstructure:"type" validate:"required"`
 	Enabled           bool           `mapstructure:"enabled"`
 	DefinitionVersion int64          `mapstructure:"definition_version" validate:"required"`
--- a/cli/internal/providers/destination/rule_spec_syntax_valid_test.go
+++ b/cli/internal/providers/destination/rule_spec_syntax_valid_test.go
@@ -1,6 +1,7 @@
 package destination
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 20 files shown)


cli/internal/providers/destination/rule_spec_syntax_valid_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}, results)
+}
+
+func TestSpecSyntaxValidRuleRequiredFields(t *testing.T) {
+	t.Parallel()
+
+	results := runSyntaxRule(t, ruleTestRegistry(t), map[string]any{})
+	assert.Equal(t, []vrules.ValidationResult{
+		{Reference: "/spec/id", Message: "'id' is required"},
+		{Reference: "/spec/display_name", Message: "'display_name' is required"},
```
Jayachand, line 125: UI follows `/^[\w .-]{2,100}$/` regex on display_name

cli/internal/providers/destination/model.go
```diff
@@ -8,11 +8,11 @@ import (
...
-	DisplayName       string         `mapstructure:"display_name"`
-	Type              string         `mapstructure:"type"`
+	ID                string         `mapstructure:"id" validate:"required"`
+	DisplayName       string         `mapstructure:"display_name" validate:"required"`
+	Type              string         `mapstructure:"type" validate:"required"`
 	Enabled           bool           `mapstructure:"enabled"`
-	DefinitionVersion int64          `mapstructure:"definition_version"`
+	DefinitionVersion int64          `mapstructure:"definition_version" validate:"required"`
 	Transformation    string         `mapstructure:"transformation"` // scalar "#transformation:<id>" — object form deferred
 	Config            map[string]any `mapstructure:"config"`
```
Jayachand, line 17: destination config support `{{ path || fallback }}` and `env.X ` syntax values. Is this supported in iac ? or properly restricted ?

Jayachand (approved):

cli/internal/providers/destination/model.go
abhimanyubabbar: Would we error receive the values as env.X and {{ path || fallback }} from upstream ?

Jayachand: Got it. config-backend service do that validation. Iac passes them as it is.

abhimanyubabbar:
Addressed the review feedback in commit `5f972c45`.

### 1. `display_name` UI regex
- Added `destination_display_name` pattern matching the UI rule: `/^[\w .-]{2,100}$/`
- Applied via `validate:"required,pattern=destination_display_name"` on `DestinationSpec.display_name`
- Added syntax-rule tests for valid, too-short, invalid-character, and too-long names

### 2. Config dynamic values (`env.X`, `{{ path || fallback }}`)
- Aligned with `terraform-provider-rudderstack`: UI dynamic values are **allowed** as opaque pass-through config values (not rejected)
- Added `IsDynamicConfigValue()` for:
  - `env.VAR`
  - `{{ path || fallback }}`
  - IaC `{{ .VAR }}` / `{{ .VAR | default }}`
- Added `dynamic_or_oneof` validator so enum fields (e.g. `connection_mode`, `types_of_client`) accept either literal values or dynamic syntax
- Added unit + syntax-rule tests covering env refs, UI template fallback, and IaC variable substitution

PR #666, opened 2026-07-09
feat(import): core smart-merge machinery behind importMerge flag

Jayachand:
## 🔗 Ticket

resolves [DEX-478](https://linear.app/rudderstack/issue/DEX-478/smart-import-merge-core-matcher-machinery-behind-importmerge-flag)

---

## Summary

First PR of the incremental smart-import (merge) delivery. When a workspace has diverged from the local project, `import workspace` today either blocks or generates duplicate specs for resources that already exist locally. This PR lands the core conflict-detection machinery for `import workspace --merge`: a per-resource-type matcher framework that links unmanaged remote resources to matching local project resources (manifest entry only, no duplicate spec). **Zero behavior change** — no provider registers matchers yet, and the `--merge` flag is rejected unless the `importMerge` experimental flag is enabled. Follow-up PRs add matchers per provider (transformations → datacatalog → event-stream → retl → datagraph). Design spec: `docs/superpowers/specs/2026-06-30-smart-import-lld-v2.md` (two design-review rounds).

---

## Changes

* New `cli/internal/provider/importmatcher` package: `Scope` (local graph + remote graph + importable), per-type `Matcher` registration, `Mark` — deterministic first-wins marking in place (sorted remotes, claimed-map with `Warn` on duplicate claims), `ByData`/`ByRawData` lookup helpers, trailing-ID reference rewrite preserving provider-specific reference shapes
* `RemoteResource.MatchedWith *Resource` — entity reference (not a boolean) so future `--replace`/interactive flows know *what* matched
* `ResourceMatcherProvider` added as a required member of the `Provider` interface (declared in `provider/provider.go` alongside the other sub-interfaces); `EmptyProvider` carries the nil default so every provider compiles unchanged — empty matchers ⇒ provider does not support smart import
* `CompositeProvider.ResourceMatchers()` aggregation (plain iteration; cross-provider order immaterial — resource types are disjoint and matchers never read another provider's marks)
* `importMerge` experimental flag; `--merge` on `import workspace` rejected in `PreRunE` when the flag is off
* Sync check under merge relaxed: only pending deletions block (`checkSyncStatus`); without `--merge`, behavior unchanged
* Importer runs `markMatchedWith` after `LoadImportable`, before reference resolution — matched resources stay in the importable collection so cross-resource references resolve

---

## Testing

[...]

PR #667, opened 2026-07-09
feat(transformations): import-merge matchers for libraries and transformations

Jayachand:
## 🔗 Ticket

resolves [DEX-479](https://linear.app/rudderstack/issue/DEX-479/smart-import-merge-transformations-matchers-library-transformation)

---

## Summary

PR2 of the incremental smart-import (merge) delivery, **stacked on #666** (core machinery). Transformations becomes the first provider to opt into `import workspace --merge` conflict detection: unmanaged remote libraries and transformations that match existing local project resources are linked (manifest entry with the local URN) instead of exported as duplicate specs. Also lands the shared `BaseHandler.FormatForExport` partition every BaseHandler-based provider will reuse. Still fully gated behind the `importMerge` experimental flag; providers without matchers remain untouched.

---

## Changes

* Transformations provider overrides `ResourceMatchers()` (replacing the `EmptyProvider` nil default): **libraries match on `import_name`** — the unique handle used in transformation code (display names are not unique; `Metadata().Name` is deliberately not used) — and **transformations match on `name`**; libraries registered first, mirroring handler registration order. Empty match fields never link.
* `BaseHandler.FormatForExport` now partitions matched vs unmatched resources: matched resources contribute a manifest entry only (emitted by BaseHandler, which has `WorkspaceID` via `Metadata()` — the `MatchedWith` reference lives on the wrapper the Impl never sees), unmatched resources flow to `Impl.FormatForExport` exactly as today. Disjoint sets, one entry emitter each — no double-emit, and no spec file for matched resources. All-matched ⇒ entries only, zero entities.

---

## Testing

* Unit (TDD, red-green): partition — matched resource ⇒ entry present/spec absent, unmatched ⇒ both, all-matched ⇒ entries only (via the example writer fixture); library matcher (import-name match, mismatch, empty-guard); transformation matcher (name match, mismatch, empty-guard); matcher registration order

---

## Risk / Impact

Low — the partition is a no-op when nothing is matched (the only way anything gets matched is `--merge` behind the experimental flag), and the matchers are pure functions reading the local graph. All other BaseHandler providers keep the nil default.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


cli/internal/providers/transformations/matchers.go
```diff
@@ -0,0 +1,46 @@
...
+
+func matchLibrary(scope importmatcher.Scope, r *resources.RemoteResource) *resources.Resource {
+	remote, ok := r.Data.(*model.RemoteLibrary)
+	if !ok || remote.ImportName == "" {
+		return nil
+	}
+
+	local, _ := importmatcher.ByRawData(scope.LocalGraph, ttypes.LibraryResourceType, func(raw any) bool {
+		library, ok := raw.(*model.LibraryResource)
+		return ok && library.ImportName == remote.ImportName
```
abhimanyubabbar: Should we remove `ok` from the type assertion as it can hide failures ?

cli/internal/providers/transformations/matchers.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+	return local
+}
+
+func matchTransformation(scope importmatcher.Scope, r *resources.RemoteResource) *resources.Resource {
+	remote, ok := r.Data.(*model.RemoteTransformation)
+	if !ok || remote.Name == "" {
+		return nil
+	}
```
abhimanyubabbar: Let's remove these silent failure points which can cause cause issues in terms of matching.

abhimanyubabbar (approved):

PR #668, opened 2026-07-10
feat(datacatalog): import-merge matchers for catalog resources

Jayachand:
## 🔗 Ticket

resolves [DEX-480](https://linear.app/rudderstack/issue/DEX-480/smart-import-merge-datacatalog-matchers-category-event-property-custom)

---

## Summary

PR3 of the incremental smart-import (merge) delivery, **stacked on #667** (transformations) → #666 (core). Datacatalog opts into `import workspace --merge` conflict detection for all five resource types. Match keys were revalidated against the provider's own in-code uniqueness rules before implementation: category/custom-type/tracking-plan are unique by `name` (`countMap[name]`), events by `name`+`eventType` (rule key `name|type`; non-track events carry empty names, enforced by syntax validation), properties by `name`+`type`+`itemTypes` (rule message "duplicate name, type and itemTypes"). Still fully gated behind the `importMerge` experimental flag.

---

## Changes

* Datacatalog provider overrides `ResourceMatchers()` with five matchers (dependency order: category, custom-type, property, event, tracking-plan); empty match fields never link
* Property matcher guards, all verified against code: remote properties linked to a custom type (`DefinitionId != ""`) never smart-link; local custom-type references never smart-link (their `type` is a `resources.PropertyRef`, not a string); custom-type **item** references never smart-link; `item_types` compared order-insensitively across the key-spelling split (local `item_types` snake_case/pre-sorted vs remote `itemTypes` camelCase/unsorted); multi-types compare as the sorted comma-joined string both sides already normalize to
* `MatchedWith` skip in all five importremote `FormatForExport`s: matched resources produce a manifest entry only — **no spec content and no spec-embedded import metadata** (the written file doesn't contain them, so its embedded `ImportIds` must not claim them); all-matched ⇒ entries only, no spec file. Tracking plans write one file per plan, so a matched plan writes nothing.

---

## Testing

* Unit (TDD, red-green): matcher registration; per-matcher predicates incl. event non-track empty-name match, property custom-type guards (local ref, remote DefinitionId, item refs), cross-spelling item-types match/mismatch, multi-type equality; export skip per resource type — matched absent from spec content, present in manifest entries under the adopted local URN; all-matched ⇒ entries only
* Full unit suite green; `make lint` 0 issues
* Behavior unchanged without the `importMerge` flag + `--merge`

---

## Risk / Impact

Low — matchers are pure local-graph lookups reachable only via `--merge` behind the experimental flag; the export skip is a no-op when nothing is matched.

---

## Checklist

[...]

PR #669, opened 2026-07-10
feat(event-stream): import-merge matcher for sources

Jayachand:
## 🔗 Ticket

resolves [DEX-481](https://linear.app/rudderstack/issue/DEX-481/smart-import-merge-event-stream-source-matcher)

---

## Summary

PR4 of the incremental smart-import (merge) delivery, a **sibling of #667 and #668 stacked on #666** (core). Event-stream sources opt into `import workspace --merge` conflict detection. Depends only on the core machinery — no dependency on the BaseHandler partition or datacatalog, so it merges independently. Still fully gated behind the `importMerge` experimental flag.

---

## Changes

* Source matcher (in the `source` package, co-located with its remote type and `NameKey`) matching on `name`, mirroring the source uniqueness rule (`countMap[spec.Name]`); registered on the event-stream provider via `ResourceMatchers()`, replacing the `EmptyProvider` nil default. Empty name never links.
* `MatchedWith` skip in the source handler's `FormatForExport`: sources write one spec file each, so a matched source appends its manifest entry (under the adopted local URN) and skips spec-file generation. The per-source embedded import metadata is only built for unmatched sources.

---

## Testing

* Unit (TDD, red-green): matcher (name match, mismatch, empty-guard); provider registration; export skip — matched source absent from written files, present in manifest entries; unmatched source still exported

---

## Risk / Impact

Low — a pure local-graph matcher reachable only via `--merge` behind the experimental flag; the export skip is a no-op when nothing is matched.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


cli/internal/providers/event-stream/source/matcher.go
```diff
@@ -0,0 +1,31 @@
...
+
+func matchSource(scope importmatcher.Scope, r *resources.RemoteResource) *resources.Resource {
+	// Dispatched by resource type, so a wrong payload is a wiring bug — panic.
+	remote := r.Data.(*sourceClient.EventStreamSource)
+	if remote.Name == "" {
+		return nil
+	}
+
+	local, _ := importmatcher.ByData(scope.LocalGraph, ResourceType, func(data resources.ResourceData) bool {
+		name, _ := data[NameKey].(string)
```
abhimanyubabbar: Given name is always a string, panic here as well if not rather than silently failing.

abhimanyubabbar (approved):
Nit about the silent failure.

PR #670, opened 2026-07-10
feat(retl): import-merge matcher for sql models

Jayachand:
## 🔗 Ticket

resolves [DEX-482](https://linear.app/rudderstack/issue/DEX-482/smart-import-merge-retl-sql-model-matcher)

---

## Summary

PR5 of the incremental smart-import (merge) delivery, a **sibling of #667/#668/#669 stacked on #666** (core). RETL SQL models opt into `import workspace --merge` conflict detection. Depends only on the core machinery, so it merges independently. Still fully gated behind the `importMerge` experimental flag.

---

## Changes

* SQL model matcher (in the `sqlmodel` package, co-located with its keys and remote type) matching on `display_name` + `account_id`; registered on the retl provider via `ResourceMatchers()`, replacing the `EmptyProvider` nil default. Empty display name never links.
* `MatchedWith` skip in the sqlmodel handler's `FormatForExport`: models write one spec file each, so a matched model appends its manifest entry (under the adopted local URN) and skips spec-file generation.

---

## Design note — match key is stricter than the uniqueness rule (intentional)

The local uniqueness rule keys on `display_name` **alone** (`countMap[displayName]`). The matcher additionally compares `account_id`, because a SQL model belongs to an account and two accounts can legitimately hold same-named models. Matching on `display_name` alone risks a **false link** across accounts; the extra key avoids that. When it doesn't link, the resource falls back to the namer (new spec, user-reconcilable) — the safe direction. Flagged here so a reviewer can confirm the product intent (per-account vs per-workspace model uniqueness upstream).

---

## Testing

* Unit (TDD, red-green): matcher (display_name+account_id match, name mismatch, account mismatch, empty-guard); provider registration; export skip — matched model absent from written files, present in manifest entries; unmatched model still exported

---

## Risk / Impact

Low — a pure local-graph matcher reachable only via `--merge` behind the experimental flag; the export skip is a no-op when nothing is matched.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


abhimanyubabbar (approved):
Approving but with ability to make sure we don't silently pass on the cases which shouldn't happen.

PR #673, opened 2026-07-13
feat(cli): gate import-manifest behind ImportMerge flag

abhimanyubabbar:
## 🔗 Ticket

Resolves [DEX-483](https://linear.app/rudderstack/issue/DEX-483/gate-the-import-manifest-through-the-experimental-flag)

---

## Summary

Gates `import-manifest.yaml` generation and recognition behind a new `ImportMerge` experimental flag so the unfinished feature stays inert by default. Without the flag, `import workspace` no longer writes the manifest, and any leftover/hand-written `import-manifest` kind fails syntax validation as an unsupported kind.

---

## Changes

- Add `ImportMerge` experimental flag (`importMerge` / `RUDDERSTACK_X_IMPORT_MERGE`)
- Gate import-manifest file generation in `WorkspaceImport`
- Gate import-manifest kind/patterns/rules (and related docs fragments) in `BuildRegistry` / `ruledoc.Build`
- Update tests to enable the flag where the feature is exercised; add coverage for flag-off behavior

---

## Testing

- Unit tests for importer flag on/off generation
- Unit tests for BuildRegistry known-kind inclusion/rejection
- Updated project/ruledoc wiring tests
- `make lint` and `make test` passed

---

## Risk / Impact

Low
Notes: default behavior removes unfinished artifact generation; enabling the experimental flag restores previous behavior.

---

## Checklist

- [ ] Ticket linked
- [x] Tests added/updated


cli/internal/project/project.go
```diff
@@ -354,13 +355,19 @@ func (p *project) registry() (rules.Registry, error) {
...
+// leftover/hand-written manifest files fail syntax validation.
 func BuildRegistry(provider, manifestProvider ProjectProvider) (rules.Registry, error) {
 	resourcePatterns := provider.SupportedMatchPatterns()
-	manifestPatterns := manifestProvider.SupportedMatchPatterns()
 
-	// Union of both providers' patterns: the source of truth for which kinds and
-	// versions the validation pipeline treats as known.
-	activePatterns := append(append([]rules.MatchPattern{}, resourcePatterns...), manifestPatterns...)
+	activePatterns := append([]rules.MatchPattern{}, resourcePatterns...)
+	importMergeEnabled := config.GetConfig().ExperimentalFlags.ImportMerge
```
sivashanmukh: `BuildRegistry` is otherwise a pure function of the two providers it's handed — reading `config.GetConfig()` here couples the builder to global mutable state and forces every test to flip the global flag to exercise either path. Suggest resolving the flag in the two callers (`project.registry()` and `deps.Registry()`) and passing it in as a parameter (`importMergeEnabled bool`), keeping the builder testable in isolation. The same global is also read independently in `ruledoc.go`; a param removes that second read.

cli/internal/ruledoc/ruledoc.go
```diff
@@ -43,14 +44,24 @@ func Build(cp provider.Provider, cliVersion, generatedAt string) (docs.Documente
...
+		entries = append(entries, manifestProvider.RuleDocEntries()...)
+	}
 	projectEntries, err := docs.LoadRuleDocEntries(projectdocs.FragmentsFS, ".")
 	if err != nil {
 		return docs.DocumentedRules{}, nil, fmt.Errorf("loading project rule docs: %w", err)
 	}
-	entries = append(entries, projectEntries...)
+	for _, e := range projectEntries {
+		// manifest-inline-conflict only registers when importMerge is on.
+		if !importMergeEnabled && e.RuleID == "project/manifest-inline-conflict" {
```
sivashanmukh: Stringly-typed coupling: matching on `e.RuleID == "project/manifest-inline-conflict"` silently stops filtering if the rule is ever renamed, leaking the fragment into gated docs. The rule and its doc fragment should ride one gate — move the `manifest-inline-conflict` fragment under the manifest provider's docs FS (already gated via `manifestProvider.RuleDocEntries()`), so no rule-ID string match is needed here.

cli/internal/project/project.go
```diff
@@ -354,13 +355,19 @@ func (p *project) registry() (rules.Registry, error) {
 // gatekeeper rules treat both resource kinds and import-manifest as known. The
 // resource-scoped gatekeepers (metadata-syntax-valid, duplicate-urn) are scoped
 // to resourcePatterns alone so the engine never hands them a manifest spec.
+//
+// Import-manifest patterns and dedicated rules are only included when the
+// importMerge experimental flag is enabled — otherwise the kind is unknown and
+// leftover/hand-written manifest files fail syntax validation.
 func BuildRegistry(provider, manifestProvider ProjectProvider) (rules.Registry, error) {
```
sivashanmukh: Maintainability: the gate is now 4 scattered `if importMergeEnabled` branches here plus 2 in `ruledoc.go`. Any new manifest contribution needs a matching branch or it silently leaks past the gate. Worth considering the "disabled provider advertises nothing" approach, which collapses all of these to a single gate on the provider.

sivashanmukh (approved):

cli/internal/project/importer/importer.go
```diff
@@ -104,13 +97,22 @@ func WorkspaceImport(
 		return fmt.Errorf("writing files for formattable entities: %w", err)
 	}
 
-	if manifestNode != nil {
-		manifestEntity := writer.FormattableEntity{
-			Content:      manifestNode,
-			RelativePath: importmanifest.FileName,
+	// Only emit the import-manifest when the importMerge experimental flag is
+	// enabled — the feature is incomplete and the artifact would confuse users.
+	if config.GetConfig().ExperimentalFlags.ImportMerge {
```
Jayachand, line 102: The visible section to user is having new import manifest yaml. Having experimental flag on write path would be minimal change. Flag can omitted on Validation rules given user do not have idea on manifest yamls.

Jayachand (approved):

cli/internal/project/importer/importer.go
abhimanyubabbar: Given open source repo, it should be assumed that we block the process from multiple sides. My reason was that only

PR #676, opened 2026-07-14
feat: add base support for s3 destination

abhimanyubabbar:
## 🔗 Ticket

Resolves [DEX-470](https://linear.app/rudderstack/issue/DEX-470/onboard-s3-and-ga4-destination-in-cli) (S3 portion; GA4 tracked separately in [DEX-471](https://linear.app/rudderstack/issue/DEX-471/onboard-ga4-destination-definition))

---

## Summary

Onboards the classic Amazon S3 destination as a manageable resource in the CLI's destination provider, and lands the syntactic/semantic validation rules, consent-management overrides, and dynamic-value handling needed to support it (and future destination definitions) end-to-end.

---

## Changes

- Add `s3` destination definition (`definitions/s3`) mapping local YAML config to the API config, mirroring `terraform-provider-rudderstack`'s `destination_s3.go` field set and `access_key_id`/`access_key` as secret keys
- Extend `DestinationDefinition`/`RegisteredDefinition` with `APIType`, `ConsentValidationOverrides`, and a registry reverse-index (`GetByAPIType`) keyed by `(APIType, Version)` for remote/import lookups
- Add `consent_validation.go` to validate `consent_management` blocks per source type, with per-destination override support
- Add `source_type_mapping.go` (common) plus a `dynamic_or_oneof` validator tag (`dynamicvalues.go`) to accept RudderStack UI dynamic syntax (`env.VAR`, `{{ path || fallback }}`) and IaC variable substitution (`{{ .VAR }}`) as pass-through values
- Wire up `SyntacticRules`/`SemanticRules`/`RuleDocEntries` on the destination provider (previously stubbed out) via new `rule_spec_syntax_valid.go` and `rule_semantic_valid.go`, with embedded rule docs
- Add `destination_display_name` validator pattern (`constants.go`) mirroring the UI's display-name constraint
- Extend `handler.go`/`model.go` for the new definition fields and update `unknownkeys.go`/`validator.go` accordingly

---

## Testing

- Unit tests added/updated across `definitions`, `definitions/common`, `definitions/s3`, and the destination provider/handler (`make test` equivalent: `go test ./cli/internal/providers/destination/...` — all passing)
- `go build ./...` verified clean

---

## Risk / Impact

Low — additive: new destination type and previously-stubbed validation rules for the destination provider. No changes to existing destination types' behavior.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


abhimanyubabbar:
@cursoragent review

cli/internal/providers/destination/definitions/common/consent_management.go
```diff
@@ -19,10 +16,13 @@ func Properties(sourceTypes []string) []converter.ConfigProperty {
 
 	properties := make([]converter.ConfigProperty, 0, len(sourceTypes))
 	for _, sourceType := range sourceTypes {
-		localSourceType := camelToSnake(sourceType)
+		remoteSourceType, ok := apiSourceType(sourceType)
```
abhimanyubabbar: Should we use the naming as localSourceType instead of sourceType to better differentiate between the two ?

cli/internal/app/dependencies.go
```diff
@@ -214,10 +215,10 @@ func setupProviders(c *client.Client) (*Providers, error) {
 	wsp := workspace.New(c)
 	dgp := dgProvider.NewProvider(dgClient.NewRudderDataGraphClient(c), c.Accounts)
 
-	// Destination registry ships empty this ticket — production destination
-	// definitions (webhook/ga4/s3) are onboarded in a follow-up. The handler
-	// consumes whatever definitions the registry holds at construction time.
 	destRegistry := definitions.NewRegistry()
+	if err := destRegistry.Register(s3.NewDefinition()); err != nil {
```
abhimanyubabbar: Should we add a feature flag here to only register the definitions if we have the experimental flag on ?

If the registry never gets populated, we will not see them in import / export layers. 

Maybe name the feature flag as DestinationSupport ?

cli/internal/providers/destination/definitions/common/consent_management_test.go
```diff
@@ -54,50 +53,84 @@ func TestPropertiesSingleSourceType(t *testing.T) {
...
+		"cloud_source":   "cloudSource",
+		"cordova":        "cordova",
+		"flutter":        "flutter",
+		"ios":            "ios",
+		"ios_swift":      "iosSwift",
+		"react_native":   "reactnative",
+		"shopify":        "shopify",
+		"unity":          "unity",
+		"warehouse":      "warehouse",
+		"web":            "web",
```
abhimanyubabbar: Why not use this list from the source_type_mapping itself rather than us defining it again ?

cli/internal/providers/destination/definitions/common/consent_management_test.go
```diff
@@ -54,50 +53,84 @@ func TestPropertiesSingleSourceType(t *testing.T) {
...
-	require.True(t, ok)
-	assert.Contains(t, properties, "web")
-	assert.Contains(t, properties, "react_native")
+	back, err := converter.APIToLocal(props, expectedAPI)
+	require.NoError(t, err)
+	assert.Equal(t, local, back)
 }
 
-func TestMergeSchemas(t *testing.T) {
+func TestPropertiesMultipleSourceTypes(t *testing.T) {
```
abhimanyubabbar: This function doesn't test anything, let's remove it ?

cli/internal/providers/destination/definitions/common/consent_validation.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"fmt"
+	"regexp"
+)
+
+var consentValuePattern = regexp.MustCompile(`^(\{\{.*\|\|.*\}\}|env\..+|.{0,100})$`)
+
+// ConsentManagement stores consent entries by local source type.
+type ConsentManagement map[string][]ConsentEntry
```
abhimanyubabbar, line 11: Let's write a README.md file in the common directory which explains the consent management and what the structure in YAML would look like for multiple source types through a snippet.

In addition to this, it explains the validations on the consent types and consent entries.

cli/internal/providers/destination/definitions/common/consent_validation.go
```diff
@@ -0,0 +1,102 @@
...
+
+func validateConsentValues(entryIndex int, consents []string) []ConsentValidationError {
+	var errors []ConsentValidationError
+	for consentIndex, consent := range consents {
+		if consentValuePattern.MatchString(consent) {
+			continue
+		}
+		errors = append(errors, ConsentValidationError{
+			Path:    fmt.Sprintf("/%d/consents/%d", entryIndex, consentIndex),
+			Message: "'consent' must be at most 100 characters or use template/environment syntax",
```
abhimanyubabbar: let's keep the message to be `'consent' must be at most 100 characters`

cli/internal/providers/destination/definitions/common/consent_validation.go
```diff
@@ -0,0 +1,102 @@
...
+		})
+	}
+	return errors
+}
+
+func isValidConsentProvider(provider *string) bool {
+	if provider == nil {
+		return true
+	}
```
abhimanyubabbar: Let's block the provider to be always required ?
I mean under what circumstances would we have a provider as nil ?

cli/internal/providers/destination/definitions/common/consent_validation.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+var consentValuePattern = regexp.MustCompile(`^(\{\{.*\|\|.*\}\}|env\..+|.{0,100})$`)
+
+// ConsentManagement stores consent entries by local source type.
+type ConsentManagement map[string][]ConsentEntry
+
+// ConsentEntry is one consent-provider configuration.
+type ConsentEntry struct {
+	Provider           *string  `mapstructure:"provider"`
```
abhimanyubabbar: Let's make Provider as string meaning if we have ConsentEntry defined for a sourcetype, it should very well have a Provider defined.

cli/internal/providers/destination/definitions/common/source_type_mapping.go
```diff
@@ -0,0 +1,52 @@
...
+func apiSourceType(typ string) (string, bool) {
+	apiSourceType, ok := apiSourceTypeByLocal[typ]
+	return apiSourceType, ok
+}
+
+// ValidateSourceTypes verifies that every local source type has an API config key.
+func ValidateSourceTypes(sourceTypes []string) error {
+	for _, sourceType := range sourceTypes {
+		if _, ok := apiSourceType(sourceType); !ok {
+			return fmt.Errorf("source type %q has no API mapping", sourceType)
```
abhimanyubabbar:
```suggestion
			return fmt.Errorf("Local source type %q has no API mapping", sourceType)
```

cli/internal/providers/destination/definitions/definition.go
```diff
@@ -87,9 +83,12 @@ func (d *RegisteredDefinition) ConnectionModes(sourceType string) ([]string, err
 	return append([]string(nil), modes...), nil
 }
 
+// IsSourceTypeSupported reports whether sourceType is in SourceTypes
 func (d *RegisteredDefinition) IsSourceTypeSupported(sourceType string) bool {
```
abhimanyubabbar: Let's remove this function IsSourceTypeSupported given no one is using it right now ?

sivashanmukh (approved):

cli/internal/app/dependencies.go
```diff
@@ -214,23 +206,49 @@ func setupProviders(c *client.Client) (*Providers, error) {
...
+	}
+
+	return providers, providerMap, nil
+}
+
+// newDestinationRegistry builds the destination definition registry. Definitions
+// are only registered when the destinationSupport experimental flag is on.
+func newDestinationRegistry(cfg config.Config) (*definitions.Registry, error) {
+	registry := definitions.NewRegistry()
+	if !cfg.ExperimentalFlags.DestinationSupport {
```
sivashanmukh, line 245: Redundant flag check. `newDestinationRegistry` is only ever called from inside the `if cfg.ExperimentalFlags.DestinationSupport` guard in `setupProviders`, so this early-return branch is dead — yet the doc comment ("Definitions are only registered when the destinationSupport experimental flag is on") implies this function is meant to be the single gate. Suggest picking one owner: either drop this inner check, or drop the outer guard and let the function decide.

sivashanmukh (commented):
### Review focus: is this non-impactful with `RUDDERSTACK_X_DESTINATION_SUPPORT` off?

**Yes — non-impactful at runtime for anyone not using destination specs, and no nil-deref risk.**

- `setupProviders` registers the destination provider (into both `providerMap` and `providers.Destination`) only inside the `DestinationSupport` guard, so it's simply absent from the composite when the flag is off.
- `providers.Destination` is only ever *assigned*, never read — so the field being `nil` when the flag is off cannot panic.
- Known-kind resolution (`activePatterns`/`isKnownKind`/`BuildRegistry`) derives kinds from the composite provider's `SupportedMatchPatterns()`, so no other provider's kinds/rules change.

**Only behavioral delta vs. `main` (flag off):** on `main` the destination provider was registered unconditionally with an *empty* registry, so `kind: destination` is currently recognized-but-unusable; after this PR it becomes an unknown kind (silently skipped in the typer/migrate path, `ErrUnsupportedSpecKind` in validate/apply). Since destinations were never usable with an empty registry, this is a cleanup, not a regression.

Inline notes below. Overall LGTM on the gating.

cli/internal/providers/destination/handler.go
```diff
@@ -197,12 +207,13 @@ func (h *HandlerImpl) MapRemoteToState(
 		return nil, nil, fmt.Errorf("managed destination %s has empty external ID", remote.ID)
 	}
 
-	if _, err := h.registry.Get(remote.Type, remote.Version); err != nil {
+	version := int64(remote.Version)
+	registered, err := h.registry.GetByAPIType(remote.Type, version)
```
sivashanmukh, line 220: Flag-on correctness (applies to all `GetByAPIType(..., remote.Version)` / `d.Version` lookups here): `client.Destination.Version` is `json:"version,omitempty"` (int64), so it is `0` when the API omits it. The S3 definition is keyed at version 1, so a managed/importable destination whose payload lacks `version` would fail with "unregistered type ... version 0". All updated tests hardcode `"version":1` — please confirm the destinations API always returns `version`, otherwise consider normalizing `0 -> 1` (or the definition's version) on read.

.github/workflows/gen-rule-docs.yml
```diff
@@ -44,6 +44,9 @@ jobs:
       # fragment (or a fragment references an unknown rule), failing the build.
       - name: generate rule catalog
         run: make gen-rule-docs
+        env:
+          RUDDERSTACK_CLI_EXPERIMENTAL: ${{ vars.RUDDERSTACK_CLI_EXPERIMENTAL || 'true' }}
+          RUDDERSTACK_X_DESTINATION_SUPPORT: ${{ vars.RUDDERSTACK_X_DESTINATION_SUPPORT || 'true' }}
```
sivashanmukh, line 49: This defaults the flag to `'true'` when the repo variable is unset, so the generated `docs/generated/rules.yaml` artifact and the drift/completeness gate now always include destination + experimental rules regardless of the flag's production state. That's an effect independent of the runtime flag — fine and probably intended, just flagging it explicitly. (No committed file changes since this is an uploaded artifact with `persist-credentials: false`.)

PR #679, opened 2026-07-14
feat(import): allow merge with pending deletions, erroring only on conflicting references

Jayachand:
## 🔗 Ticket

resolves [DEX-485](https://linear.app/rudderstack/issue/DEX-485/smart-import-merge-relax-pending-deletions-block-to-conflict-only)

---

## Summary

Relaxes the blanket `--merge` block on pending deletions (a sibling of the provider PRs, stacked on #666). Previously any unapplied local deletion blocked the merge import wholesale. Now the import proceeds, and errors **only** when an importable resource actually references a locally-deleted managed resource — detected lazily and precisely at reference resolution, where the failure naturally occurs, instead of via pre-flight scanning (which would need new per-provider interface surface to enumerate references inside typed payloads). Still gated behind the `importMerge` experimental flag.

---

## Changes

* `checkSyncStatus`: no pre-flight block under merge — divergence, including pending deletions, is the point of merging
* `resolver.ErrPendingDeleteConflict`: the resolver's found-in-`Remote`/missing-from-`Graph` branch — which under merge can only mean an unapplied local deletion (non-merge blocks any diverged project before resolution runs) — now returns an actionable sentinel naming the deleted URN and remote ID: `reference target is deleted locally but the deletion is not applied: category:checkout (remote cat_1); apply the pending deletion or restore its spec before importing with --merge`. Handler wrapping prepends the referencing resource. The sentinel survives `%w` chains, so callers can `errors.Is` it.
* `initNamer` now preloads the **union of target + source** graph IDs: a pending-deleted resource is absent from the target graph but its remote twin still holds the ID upstream, so the ID stays reserved and an unrelated importable (normalization-collapse name) cannot mint it. No-op for non-merge (source ≈ target there).

---

## Safety

* **No partial imports**: `FormatForExport` completes before any `writer.Write`, so a conflict aborts with zero files written
* Deleted resources are managed (never in `Importable`), absent from `LocalGraph` (never matched), and never re-exported — the import cannot resurrect a deletion
* Known limitation (documented): conflicts surface one at a time — the first unresolvable reference aborts. Collecting all would require handlers to continue past resolution errors; future polish.

---

## Testing

* Unit (TDD, red-green): `checkSyncStatus` matrix updated (merge + deletions now proceeds); resolver — importable hit, pending-delete conflict carries sentinel + URN + remote ID, unknown remote is not misclassified as a conflict; namer — pending-deleted ID is not minted for a new importable

---

## Risk / Impact

Low — three localized sites (`checkSyncStatus`, resolver error, namer preload) behind the experimental flag; no matcher/export/manifest surface touched, so the sibling provider PRs are unaffected.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


Jayachand:
test pr

PR #681, opened 2026-07-15
fix(typer): resolve analytics instance lazily in generated TS client (stop silent event drops)

sivashanmukh:
### The bug (silent event loss)
The generated `RudderTyper` class captured the analytics instance in its constructor and every method called `this.analytics.*`:

```ts
constructor(analytics: RudderAnalytics) { this.analytics = analytics; }
```

In a browser app the standard RudderStack snippet installs a **buffering preloader** on `window.rudderanalytics` and swaps it for the real SDK once `analytics-js` loads. So the obvious call:

```ts
export const analytics = new RudderTyper(window.rudderanalytics); // runs at import, captures the preloader
```

captured the preloader, and every event fired **after** the SDK loaded went into an abandoned queue. No error, no type error, no failed build: events just quietly stopped arriving. Found while migrating rudder-webapp off the npm typer.

The v1 npm generator did **not** have this. It resolved `window.rudderanalytics` per call via a closure.

### The fix
The constructor is now **resolver-only**, and an `analytics` getter re-resolves on every call:

```ts
constructor(resolveAnalytics: () => RudderAnalytics) {
    this.resolveAnalytics = resolveAnalytics;
}
private get analytics(): RudderAnalytics { return this.resolveAnalytics(); }
```

Consumers pass `new RudderTyper(() => window.rudderanalytics)` (now the documented example in the generated file). Node/server usage passes `() => analytics`.

Resolver-only rather than `RudderAnalytics | (() => RudderAnalytics)`, per review feedback: accepting an instance would leave the unsafe call compiling, and the failure mode it produces is silent. A thunk is trivial to write at every call site, so the union buys nothing but a footgun.

### Breaking change
This changes the generated constructor signature, so previously generated clients need `new RudderTyper(analytics)` updated to `new RudderTyper(() => analytics)`. `rudder-cli typer` is still behind `RUDDERSTACK_CLI_EXPERIMENTAL` and pre-GA, so the blast radius is limited to the current dogfooding consumers. Landing it now is the point: after GA the eager shape is locked for every customer who adopts it.

### Test
- Regenerated the TS golden file (`make typer-typescript-update-testdata`).
- Added `TestGenerateResolvesAnalyticsLazily`, asserting the generator emits the resolver + getter, never `this.analytics = analytics;`, and never the `RudderAnalytics | (() => RudderAnalytics)` union.
- Updated the `testdata/validator` suites, which exercise the generated client against the real `@rudderstack/analytics-js` SDK. The "validate generated TypeScript against RudderStack JS SDK" check passes.
- `go test ./cli/internal/typer/...` passes.

### DIFF
--- a/cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
+++ b/cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
@@ -14,12 +14,31 @@
  * analytics.load("YOUR_WRITE_KEY", "YOUR_DATA_PLANE_URL");
  * const rudderTyper = new RudderTyper(analytics);
  * ```
+ *
+ * In a browser app that loads the SDK asynchronously, the standard RudderStack
+ * snippet swaps `window.rudderanalytics` from the buffering preloader to the real
+ * SDK once it loads. Pass a resolver function instead of the instance so every call
+ * re-resolves the current SDK — otherwise the preloader is captured at construction
+ * and events fired after the SDK loads are silently dropped:
+ *
+ * ```ts
+ * const rudderTyper = new RudderTyper(() => window.rudderanalytics);
+ * ```
  */
 export class RudderTyper {
-    private readonly analytics: RudderAnalytics;
+    private readonly resolveAnalytics: () => RudderAnalytics;
 
-    constructor(analytics: RudderAnalytics) {
-        this.analytics = analytics;
+    /**
+     * @param analytics - a RudderAnalytics instance, or a function that returns the
+     * current instance. Prefer the function form in the browser so a lazily-loaded
+     * SDK is resolved on every call rather than captured once at construction.
+     */
+    constructor(analytics: RudderAnalytics | (() => RudderAnalytics)) {
+        this.resolveAnalytics = typeof analytics === "function" ? analytics : () => analytics;
+    }
+
+    private get analytics(): RudderAnalytics {
+        return this.resolveAnalytics();
     }
 {{ range $method := .AnalyticsMethods }}
 {{- if $method.Overloads }}
--- a/cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
+++ b/cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
@@ -5,6 +5,10 @@
  * - ID: {{ .TrackingPlanID }}
  * - Version: {{ .TrackingPlanVersion }}
  *
+ * The constructor takes a resolver function that returns the current
+ * RudderAnalytics instance. It is re-invoked on every call, so a lazily-loaded
+ * SDK is always resolved fresh rather than captured once at construction.
+ *
  * @example
  * ```ts
  * import { RudderAnalytics } from "@rudderstack/analytics-js";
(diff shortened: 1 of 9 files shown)


arpl:
The lazy-resolution mechanism is right — resolver + getter re-resolves on every call, so the preloader→SDK swap is picked up. This fixes the silent-drop bug.

One gap worth closing while we're in here: **the getter re-resolves, but it still assumes the resolved instance is non-null at call time.** Method bodies call `this.analytics.track(...)`, so if the resolver yields `undefined`, that throws `Cannot read properties of undefined (reading 'track')` rather than no-op'ing.

Why this matters:

- **The v1 npm generator no-op'd here.** Every generated method was `const a = analytics(); if (a) { a.track(...) }`. So a client that fires an event before/without the SDK present silently did nothing under v1, and now throws under v2 — a behavior regression for the exact consumers this PR is trying to make whole.
- **It's a real integration, not a hypothetical.** rudder-webapp gates the SDK snippet on a write key — no key → `window.rudderanalytics` is never defined. `new RudderTyper(() => window.rudderanalytics)` would throw at the first event. That's precisely why our consumer-side wrapper is a proxy (re-resolve **and** no-op when absent) rather than a bare `() => window.rudderanalytics`. If the generated client covered both, that wrapper could go away entirely — which is the drop-in-for-everyone outcome.

Concrete suggestion — optional-chain in the method template, restoring v1 semantics:

```ts
// method body
this.analytics?.track(
    "{{ $method.EventName }}",
    props as unknown as SDKApiObject,
    this.withRudderTyperContext(options),
    callback,
);
```

(`identify`/`group`/`page` overloads get the same `?.`.) Methods already return `void`, so no-op-on-nullish is a clean semantic. Would pair with a test asserting a nullish resolve doesn't throw — happy to push that as a follow-up commit or a separate PR if you'd rather keep this one tight.

Not a blocker on the core fix — this can land as-is and the nullish case is a fast follow. Flagging now because it's the difference between "consumers still need a proxy" and "the generated client is safe on its own".

cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
```diff
@@ -14,12 +14,31 @@
...
-    private readonly analytics: RudderAnalytics;
+    private readonly resolveAnalytics: () => RudderAnalytics;
+
+    /**
+     * @param analytics - a RudderAnalytics instance, or a function that returns the
+     * current instance. Prefer the function form in the browser so a lazily-loaded
+     * SDK is resolved on every call rather than captured once at construction.
+     */
+    constructor(analytics: RudderAnalytics | (() => RudderAnalytics)) {
+        this.resolveAnalytics = typeof analytics === "function" ? analytics : () => analytics;
```
akashrpo: Enables the safe lazy form, but a user can still pass an instance, new RudderTyper(window.rudderanalytics), and hit the old bug (events silently dropped once the browser SDK swaps in). The safe path is documented but nothing blocks the unsafe one. Consider making the constructor resolver-only so the instance form cannot compile.

akashrpo (changes requested):

cli/internal/typer/generator/platforms/typescript/templates/ruddertyper.tmpl
arpl: Agreed, and done in `881a9083`. The constructor is resolver-only now: `constructor(resolveAnalytics: () => RudderAnalytics)`, with `analytics` as a private getter that calls the resolver on every access. `new RudderTyper(window.rudderanalytics)` no longer compiles, so the silent form is gone rather than just documented against.

`TestGenerateResolvesAnalyticsLazily` also asserts the generator never emits the `RudderAnalytics | (() => RudderAnalytics)` union, so we cannot drift back into accepting an instance. Updated the PR description, which still described the union form.

Taking this one over from Shanmukh to get it landed for the dogfooding release. Could you re-review when you get a chance?

akashrpo (approved):

PR #685, opened 2026-07-17
feat: addded support for source type gated path on the configproperties for a definition

abhimanyubabbar:
Scanned-by: gitleaks 8.30.0

## 🔗 Ticket

Resolves [DEX-537](https://linear.app/rudderstack/issue/DEX-537/handle-gated-key-paths-by-source-types-on-the-destination-definition)

---

## Summary

Registering of config-properties on the destination definition now has ability to be gated by the source types meaning they only appear when destination is connected to  a valid source type. 

This relationship needs to be showcased in the destination definition in the registry.

### DIFF
--- a/cli/internal/providers/destination/definitions/gatedkeys.go
+++ b/cli/internal/providers/destination/definitions/gatedkeys.go
@@ -1,0 +1,75 @@
+package definitions
+
+import (
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+)
+
+// buildGatedKeyPaths derives the reverse index of source-type-gated config
+// properties (keypath -> entitled source types), failing on definitions that
+// gate unsupported source types or keys absent from the config model.
+func buildGatedKeyPaths(def *DestinationDefinition, configType reflect.Type) (map[string][]string, error) {
+	index := make(map[string][]string)
+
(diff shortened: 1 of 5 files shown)


cli/internal/providers/destination/definitions/gatedkeys.go
```diff
@@ -0,0 +1,75 @@
...
+			return nil, fmt.Errorf("duplicate gated config key %q", prop.LocalKey)
+		}
+		index[keyPath] = append([]string(nil), prop.SourceTypes...)
+	}
+
+	return index, nil
+}
+
+// localKeyToPointer converts a gjson dot path over local keys to the JSON
+// pointer form used by ConfigError paths ("a.b" -> "/a/b").
```
Jayachand: below function produces `("a.b" -> "a/b")` not intended as in comment `("a.b" -> "/a/b")`

Jayachand (approved):

cli/internal/providers/destination/definitions/gatedkeys.go
abhimanyubabbar: Yes, I updated the function but comment was old, will fix it.

dhawal1248 (approved):

PR #688, opened 2026-07-19
feat(cli): onboard attentive_tag destination definition

abhimanyubabbar:
## 🔗 Ticket

Resolves [DEX-491](https://linear.app/rudderstack/issue/DEX-491)

---

## Summary

Onboards the Attentive Tag destination (`attentive_tag` / `ATTENTIVE_TAG`) into the CLI destination definition registry so it can be managed via YAML specs when destination support is experimentally enabled.

Rebased onto current `main` and re-run through the `onboard-cli-destination` skill: the definition is now registered behind the `unverifiedDestinations` flag introduced after this branch was first written, `SkipZeroValue` filters were dropped per current converter guidance, and destination e2e fixtures + snapshots were added.

---

## Changes

- Add `definitions/attentive_tag` with config model, terraform-ported property mappings, source types, connection modes, and consent management
- Register the definition in `newDestinationRegistry` inside the `UnverifiedDestinations` block (alongside `http`, `rs`, `s3`) and extend the flag-matrix test expectation
- Port `converter.Simple` mappings **bare** — no `SkipZeroValue`, so explicit zero values (`sign_up_source_id: ""`, `enable_new_identify_flow: false`) round-trip instead of being dropped and re-appearing as phantom diffs on every plan
- Enforce strict digits-only validation on `sign_up_source_id` via the named `attentive_tag_sign_up_source_id` pattern (no dynamic-value escape hatch)
- Add destination e2e create/update fixtures, expected upstream snapshots, a dummy `ATTENTIVE_TAG_API_KEY` var, and the matching raw-secret leak assertion

### Example YAML

```yaml
version: rudder/v1
kind: destination
metadata:
  name: my-attentive-tag-destination
spec:
  id: my-attentive-tag-destination
  display_name: My Attentive Tag Destination
  type: attentive_tag
  enabled: true
  definition_version: 1
  config:
    api_key: your-attentive-api-key
    sign_up_source_id: "123456"
    enable_new_identify_flow: true
```

### Schema / source discrepancies

| Item | Detail |
| --- | --- |
| Dropped source types | `amp`, `shopify`, `warehouse` present in integrations-config `db-config.json` but omitted (S3 / CLI event-stream ownership precedent) |
| TF vs db-config sources | Terraform `supportedSourceTypes` omits `androidKotlin` / `iosSwift`; CLI keeps them from db-config |
| `sign_up_source_id` validation | Upstream schema/TF allow `env.*` and `{{ ... \|\| ... }}` in addition to `^[0-9]*$`. CLI enforces **digits only** (or omit) — intentionally stricter, no dynamic pass-through |
| `api_key` | Length constrained with `required,min=1,max=100` (matches schema `.{1,100}` after stripping dynamic prefixes) |
| Gated keys | None — `apiKey`, `signUpSourceId`, `enableNewIdentifyFlow` are all in `destConfig.defaultConfig` |
| Connection modes | All kept source types: `cloud` only (matches db-config + schema enums) |

### Config fields (supported surface)

| Local key | API key | Notes |
| --- | --- | --- |
| `api_key` | `apiKey` | required, secret (write-only upstream) |
| `sign_up_source_id` | `signUpSourceId` | optional, digits only |
| `enable_new_identify_flow` | `enableNewIdentifyFlow` | optional bool |
| `consent_management` | `consentManagement` | via `common.Properties` |

---

## Testing

- Unit tests: metadata, config validation (required `api_key`, 100-char bound, strict `sign_up_source_id` incl. rejected dynamic values, unknown keys, consent source/provider errors), and conversion round-trip incl. an explicit zero-value case — 100% statement coverage on the new package
- E2E: `attentive-tag` create/update fixtures with matching upstream snapshots. Attentive Tag has a single config shape (one auth mode, no mutually-exclusive branches), so one variation covers it; the update mutates `sign_up_source_id`, `enable_new_identify_flow`, and the display name to exercise the live update path
- Both fixture sets verified with `rudder-cli validate` (with var substitution); the example YAML above verified the same way and pinned by a `ValidateConfig` test case
- `go build ./...`, `make test`, `make lint` — all clean
- `go test ./cli/tests -run TestDestinationsApply -count=1` and `go test ./cli/tests/helpers -run TestDestinationSnapshotTester -count=1` pass (the former compiles and skips as designed)

**Snapshot caveat:** the gated live run (`RUN_DESTINATION_E2E=1`) was **not** executed — it destroys the target workspace and needs a live destination-enabled stack. The committed snapshots are derived, not captured: the config payload comes from the converter mapping, `apiKey` is omitted because it is a write-only secret, and `enableNewIdentifyFlow: false` comes from the `schema.json` default. This derivation was validated against the existing `s3`/`http` snapshots, where the schema defaults (`enableSSE`, `isBatchingEnabled`, `isDefaultMapping`, `eventFilteringOption`) exactly account for the keys the backend adds. Worth one gated run against a live stack before merge to confirm.

---

## Risk / Impact

Low

Notes: double-gated behind `experimental: true` + `flags.destinationSupport: true` + `flags.unverifiedDestinations: true`. No change to default CLI behavior when the flags are off.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


cli/internal/providers/destination/definitions/attentive_tag/definition.go
```diff
@@ -0,0 +1,78 @@
...
+	common.SourceTypeWeb,
+	common.SourceTypeUnity,
+	common.SourceTypeReactNative,
+	common.SourceTypeFlutter,
+	common.SourceTypeCordova,
+	common.SourceTypeCloud,
+}
+
+var connectionModes = map[string][]string{
+	common.SourceTypeAndroid:       {"cloud"},
```
dhawal1248, line 35: can we derive the "cloud" literal also from the common library..we support only three modes - cloud, device and hybrid so would be good to constrain these values and use constants so managing them becomes easier

dhawal1248 (dismissed):

Jayachand (approved):

PR #710, opened 2026-07-23
feat(cli): add accounts provider (map-config, shared secret helpers)

sivashanmukh:
## 🔗 Ticket

Resolves [CFD-61](https://linear.app/rudderstack/issue/CFD-61/accounts-full-implementation)

---

## Summary

Adds a declarative `account` provider so RETL accounts can be managed as YAML specs through the standard apply cycle (create / update / import / export / destroy), gated behind the experimental `AccountSupport` flag.

Account config is authored as a single flat `map[string]any` and split at the API boundary into the account definition's non-secret `options` and write-only `secret` payloads — the same map-based secret model destinations use, so no per-provider reflection or struct-tag machinery is needed. The map-config secret helpers are promoted out of the destination handler into `cli/internal/secret` and shared by both providers.

The first supported definition is `SOURCE_BIGQUERY`; the secret-key set is currently a single documented map, with the control-plane account-definitions API named as the seam for adding Snowflake/Postgres with zero per-definition code.

---

## Changes

- **New `accounts` provider** (`cli/internal/providers/accounts/`): `kind: account`, flat `config` map, per-definition secret keys, full CRUD + import/export. `splitConfig` partitions the flat config into API `options` vs `secret`.
- **Shared map-config secret helpers** (`cli/internal/secret/mapconfig.go`): `WrapKnownSecrets` / `WrapUnknownSecrets` / `RevealSecrets` / `MaskSecrets`, lifted from the destination handler; destination refactored to consume them (handler shrinks ~124 lines). Helpers keep the presence-based semantics from #709.
- **Accounts seeds its unconditional secret** before wrapping in `MapRemoteToState`, so the write-only credential (never returned by the API) is always marked unknown and always re-applied on reconcile.
- **Experimental gating**: new `AccountSupport` flag; provider registered only when enabled (mirrors `DestinationSupport`).
- **Gatekeeper rules**: `account` kind wired into the duplicate-urn / metadata-syntax-valid / manifest-inline-conflict fragments and the ruledoc coverage tests.

---

## Testing

- **Unit** (`make test`): accounts handler create/update/import/map-remote/export, including secret split, definition-immutability, remote-secret-always-unknown, and export `{{ .VAR }}` tokenization with a raw-secret leak assertion. Destination suite passes against the shared helpers, proving the refactor is behavior-preserving. Rule-catalog drift test covers the account kind.
- **E2E** (`cli/tests/command_accounts_apply_test.go`): gated behind `RUN_ACCOUNT_E2E=1` (needs a live stack with the `SOURCE_BIGQUERY` definition + PAT) — create → update → re-apply no-op, with a raw-secret leak guard on CLI output.
- `make lint` clean.

---

## Risk / Impact

Low. New provider is off by default (experimental flag). The only change to existing behavior is the destination handler consuming the relocated secret helpers, which are byte-for-byte the same presence-based logic (#709) and covered by the unchanged destination test suite.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/accounts/handler.go
+++ b/cli/internal/providers/accounts/handler.go
@@ -1,0 +1,359 @@
+package accounts
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	"github.com/rudderlabs/rudder-iac/api/client"
+	"github.com/rudderlabs/rudder-iac/cli/internal/namer"
+	"github.com/rudderlabs/rudder-iac/cli/internal/project/importmanifest"
+	"github.com/rudderlabs/rudder-iac/cli/internal/project/specs"
+	"github.com/rudderlabs/rudder-iac/cli/internal/project/writer"
+	"github.com/rudderlabs/rudder-iac/cli/internal/provider/handler"
+	"github.com/rudderlabs/rudder-iac/cli/internal/providers/transformations/handlers"
+	"github.com/rudderlabs/rudder-iac/cli/internal/resolver"
+	"github.com/rudderlabs/rudder-iac/cli/internal/resources"
+	"github.com/rudderlabs/rudder-iac/cli/internal/secret"
+)
+
+// AccountHandler is the BaseHandler instantiation for accounts.
+type AccountHandler = handler.BaseHandler[AccountSpec, AccountResource, AccountState, RemoteAccount]
+
+// HandlerMetadata describes the account handler for the framework.
+var HandlerMetadata = handler.HandlerMetadata{
+	ResourceType:     AccountResourceType,
+	SpecKind:         AccountSpecKind,
+	SpecMetadataName: AccountMetadataName,
+}
+
+// accountSecretKeys maps an account definition to its secret field set — the
+// account-side analogue of a destination definition's SecretKeys().
+//
+// ponytail: BigQuery-only, hardcoded. The real registry fetches secretFields
+// from the control-plane account-definitions API (unversioned, name-keyed).
+// Add a lookup when a second definition lands; the split logic below is already
+// definition-driven, so only this map changes.
+var accountSecretKeys = map[string][]string{
+	"SOURCE_BIGQUERY": {"credentials"},
+}
+
+// AccountStore is the subset of the accounts API client the handler needs;
+// declared at the point of use so tests inject a mock. *client.Client.Accounts
+// satisfies it.
+type AccountStore interface {
+	Create(ctx context.Context, req *client.CreateAccountRequest) (*client.Account, error)
+	Update(ctx context.Context, id string, req *client.UpdateAccountRequest) (*client.Account, error)
+	Delete(ctx context.Context, id string) error
+	Get(ctx context.Context, id string) (*client.Account, error)
+	ListAll(ctx context.Context, opts ...client.ListAccountsOption) ([]client.Account, error)
+	SetExternalID(ctx context.Context, id, externalID string) error
+}
+
+// HandlerImpl owns account CRUD against the API client.
+type HandlerImpl struct {
+	store AccountStore
+}
+
+// NewHandler builds an *AccountHandler wired to the given store.
+func NewHandler(store AccountStore) *AccountHandler {
+	return handler.NewHandler(&HandlerImpl{store: store})
+}
+
+func (h *HandlerImpl) Metadata() handler.HandlerMetadata { return HandlerMetadata }
+
+func (h *HandlerImpl) NewSpec() *AccountSpec { return &AccountSpec{} }
+
+// ExtractResourcesFromSpec resolves the account definition's secret keys and
+// wraps them in Config as *secret.String — mirrors the destination handler,
+// minus the (type, version) registry lookup (account definitions are
+// unversioned).
+func (h *HandlerImpl) ExtractResourcesFromSpec(_ string, spec *AccountSpec) (map[string]*AccountResource, error) {
+	keys, ok := accountSecretKeys[spec.AccountDefinitionName]
+	if !ok {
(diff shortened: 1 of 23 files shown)


cli/internal/providers/accounts/handler.go
```diff
@@ -0,0 +1,359 @@
...
+		AccountDefinitionName: data.AccountDefinitionName,
+		Name:                  data.ID,
+		Options:               options,
+		Secret:                secretPayload,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("creating account %q: %w", data.ID, err)
+	}
+
+	if err := h.store.SetExternalID(ctx, created.ID, data.ID); err != nil {
```
Jayachand: why not pass externalId in create call itself ?

cli/tests/command_accounts_apply_test.go
```diff
@@ -0,0 +1,71 @@
...
+			filepath.Join(projectDir, "create"), "--var-file", varFile, "--confirm=false")
+		require.NoError(t, err, "create apply failed: %s", out)
+		assert.NotContains(t, string(out), rawAccountSecret, "raw secret must never appear in CLI output")
+	})
+
+	t.Run("apply update", func(t *testing.T) {
+		out, err := executor.Execute(cliBinPath, "apply", "-l",
+			filepath.Join(projectDir, "update"), "--var-file", varFile, "--confirm=false")
+		require.NoError(t, err, "update apply failed: %s", out)
+		assert.NotContains(t, string(out), rawAccountSecret, "raw secret must never appear in CLI output")
```
Jayachand, line 64: we use verifyTestResults for e2e comparison

cli/internal/providers/accounts/model.go
```diff
@@ -0,0 +1,46 @@
...
+	Config                map[string]any `mapstructure:"config"`
+}
+
+// AccountResource is the resolved representation the differ compares. Registered
+// secret keys inside Config are wrapped as *secret.String so the differ's
+// secret-aware branch owns their comparison — identical to DestinationResource.
+type AccountResource struct {
+	ID                    string
+	AccountDefinitionName string
+	Config                map[string]any
```
Jayachand, line 30: name at same level as Config

cli/internal/providers/accounts/handler.go
sivashanmukh: Done — the create call now claims the external id inline. The api-gateway create endpoint rejected `externalId` (`Unrecognized key`), so the two-step was a backend constraint, not a choice. I've made the backend accept it and set it atomically in the same insert (config-backend rudderlabs/rudder-config-backend#6510) and the public gateway forward it (rudderlabs/rudder-control-plane#2499). `Create` now sends `ExternalID` on `CreateAccountRequest` and drops the separate `SetExternalID`. Import still uses update + set-external-id since it adopts an existing account (can't create).

cli/tests/command_accounts_apply_test.go
sivashanmukh: Reworked — the e2e now verifies upstream account state via the API + `helpers.CompareStates` against a snapshot (`testdata/expected/upstream/accounts/<dir>/account.json`), mirroring the `verifyTestResults`/`verifyState` pattern, instead of the brittle `NotContains` stdout checks. Volatile fields (id, workspaceId, timestamps) are in the ignore list; the raw-secret leak guard stays.

cli/internal/providers/accounts/model.go
sivashanmukh: Done — `Name` is now a top-level field on `AccountSpec`/`AccountResource`, a sibling of `Config`, and wired through create/update/map-remote/export. It maps to the control-plane account's display `name`; `ID` stays the external-id URN claim (they were previously collapsed — create shoved the id into both). The full GET/create/update shapes I validated against the real backend are captured in the PR.

cli/internal/providers/accounts/model.go
```diff
@@ -0,0 +1,52 @@
...
+// destination spec uses (one config map, secrets identified by key).
+//
+// ID is the caller-owned external id (the URN claim); Name is the account's
+// display name in the control plane — a distinct field, so it sits beside Config
+// rather than inside it.
+type AccountSpec struct {
+	ID                    string         `mapstructure:"id" validate:"required"`
+	Name                  string         `mapstructure:"name" validate:"required"`
+	AccountDefinitionName string         `mapstructure:"account_definition_name" validate:"required"`
+	Config                map[string]any `mapstructure:"config"`
```
Jayachand: why config is not a required ?

cli/internal/providers/accounts/handler.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// accountSecretKeys maps an account definition to its secret field set — the
+// account-side analogue of a destination definition's SecretKeys().
+//
+// ponytail: BigQuery-only, hardcoded. The real registry fetches secretFields
+// from the control-plane account-definitions API (unversioned, name-keyed).
+// Add a lookup when a second definition lands; the split logic below is already
+// definition-driven, so only this map changes.
+var accountSecretKeys = map[string][]string{
```
Jayachand: rename `accountSecretKeys` to `registeredAccountSecretKeys`? This gives
 the feel of global constant

Jayachand (approved):

cli/internal/providers/accounts/handler.go
sivashanmukh: Done — renamed to `registeredAccountSecretKeys` (var + all references).

cli/internal/providers/accounts/model.go
sivashanmukh: Good catch — added `validate:"required"` to `Config`. An account always carries options + secret, so an empty config is never valid (and the definition-driven split would produce empty options/secret and fail backend validation anyway).

PR #711, opened 2026-07-23
feat(transformations): add connect/disconnect destination endpoints

plaflamme:
## 🔗 Ticket

N/A

---

## Summary

Add client support for associating a transformation with a destination via the two dedicated API endpoints:

- POST /transformations/{id}/connectToDestination
- POST /transformations/{id}/disconnectFromDestination

Both accept a {"destinationId": "..."} body and return the updated transformation. The Transformation type gains a Destinations field so the connected destinations returned by these endpoints (and by GET /transformations/{id}) can be read back.

Adds unit tests covering success, the "not published" 400 error, and unmarshal failures.
---

## Changes

- adds connectToDestination for transformations
- adds disconnectFromDestination for transformations

---

## Testing

How was this tested?

- Unit tests

---

## Risk / Impact

Low
Notes (if any):

---

## Checklist

- [ ] Ticket linked
- [ ] Tests added/updated

### DIFF
--- a/api/client/transformations/transformations.go
+++ b/api/client/transformations/transformations.go
@@ -24,6 +24,8 @@ type TransformationStore interface {
 	ListTransformations(ctx context.Context) ([]*Transformation, error)
 	DeleteTransformation(ctx context.Context, id string) error
 	SetTransformationExternalID(ctx context.Context, id string, externalID string) error
+	ConnectToDestination(ctx context.Context, id string, req *ConnectToDestinationRequest) (*Transformation, error)
+	DisconnectFromDestination(ctx context.Context, id string, req *ConnectToDestinationRequest) (*Transformation, error)
 
 	// Library operations
 	CreateLibrary(ctx context.Context, req *CreateLibraryRequest, publish bool) (*TransformationLibrary, error)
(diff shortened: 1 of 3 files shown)


sivashanmukh:
Thanks for this! I want to make sure I understand the use-case before we decide on it.

The transformation to destination link is already supported in the CLI today. It landed with the recent destinations work via the destination-side endpoints (PUT/DELETE/GET /destinations/{id}/transformation, see api/client/destinations.go), and it's wired into the destination provider's apply cycle, so Create/Update/Delete all reconcile the link.

This PR adds the mirror endpoints from the transformation side (/transformations/{id}/connectToDestination), which aren't consumed anywhere yet. On the surface it looks like the existing destination-side flow already covers the same capability.

Could you share the workflow you're trying to enable? Specifically, is there something the current destination-side linking doesn't solve for you, or a reason you need the transformation-side direction? That'll help us confirm whether existing functionality already covers it, and it's genuinely useful for shaping where we take this repo next.

plaflamme:
@sivashanmukh I'm specifically interested in having access to "connecting" a transformation to a destination as a terraform declaration (see [this draft PR](https://github.com/rudderlabs/terraform-provider-rudderstack/pull/296)).

We have destinations that require filtering out data that should not be sent to them. As such, when creating them, we'd also like to make sure the transformation that does this filtering is applied to those destinations. We could add it manually in the UI, but this is error prone and the filtering is quite important in this case.

Our workflow currently looks like this:

1. declare a secret to hold the destination's credentials (API key or similar)
2. populate the secret manually
3. declare the destination using that secret value
4. edit the destination in the rudderstack dashboard to add the already existing transformation to it
5. connect a source to the destination

We'd like to get rid of step 4 by having a terraform declaration, i.e.:

```
resource rudderstack_transformation_connection "my_destination_filtering" {
  destinationId = rudderstack_destination_foo.my_destination.id
  transformationId = "abcd1234"
}
```

If we can achieve this some other way already, we'd be happy to use this other method, but I couldn't find how to do that in the terraform provider at the moment.

plaflamme:
@sivashanmukh is this a reasonable change to make? Or perhaps there's another approach we should take?

sivashanmukh:
@plaflamme thanks for spelling out the workflow, that's the context I needed. Dropping step 4 is reasonable.

The link is already modelled on the destination side, released in v0.22.0:

```
// api/client/destinations.go
client.Destinations.ConnectTransformation(ctx, destinationID, transformationID)  // PUT    /v2/destinations/{id}/transformation
client.Destinations.GetTransformation(ctx, destinationID)                        // GET    /v2/destinations/{id}/transformation
client.Destinations.DisconnectTransformation(ctx, destinationID)                 // DELETE /v2/destinations/{id}/transformation
```
Our experimental destination support in the CLI is built on these. There the link isn't a separate resource, it's a field on the destination spec:
```
spec:
  id: webhook-prod
  type: WEBHOOK
  transformation: "#transformation:my-transformation"
```
and the destination handler reconciles it across the apply cycle: create then connect, update diffs and re-links, delete disconnects first, import/export resolve the link back to a `#transformation:<id>` reference.

The blocker isn't the endpoints, it's that we haven't aligned internally on the Terraform shape, which is hard to walk back once people have `.tf` written against it. Two options:

1. A transformations data source plus a standalone connection resource, which is what [#296](https://github.com/rudderlabs/terraform-provider-rudderstack/pull/296) does.
2. A transformations data source plus `resource_destination` carrying the transformation ID, mirroring the CLI.

We're leaning toward (2), for consistency with the CLI and to keep one destination as one resource. Doing it properly needs the connect/disconnect to be atomic with the destination's create/update/delete, rather than a follow-up call that can leave a destination live with the wrong transformation attached. That's an API change on our end and the piece still to work out. The CLI has the same gap today.

So I'd rather not merge the transformation-side endpoints: they're a second, unconsumed path to the same link, and they'd bake in the (1)-shaped model before we've settled this.

I'll take it to the team and follow up here. If you hit something the destination-side API can't express, say so, that's useful input for the decision.

sivashanmukh:
@plaflamme To make the right decision, it would also help us to know how you currently manage transformations. Are you using one of:
1. [Legacy Transformation action](https://github.com/marketplace/actions/rudder-transformation-action)
2. [Latest CLI action](https://github.com/marketplace/actions/rudderstack-rudder-cli-project-manager)
3. Custom DIY solution
4. Something else?

plaflamme:
@sivashanmukh 

> it would also help us to know how you currently manage transformations

We're unfortunately using the UI directly at the moment. We'd use terraform when that becomes available, but in the meantime, it's been simpler for us to do this part manually.

> Two options:
>   1.  A transformations data source plus a standalone connection resource, which is what https://github.com/rudderlabs/terraform-provider-rudderstack/pull/296 does.
>   2.  A transformations data source plus resource_destination carrying the transformation ID, mirroring the CLI.

Since destinations can only have one transformation, then it would make more sense to have that be specified on the destination's declaration. A standalone connection resource allows you to declare connecting multiple transformations on a single destination; ideally you can't declare something that's not possible.

PR #713, opened 2026-07-24
fix(tests): stop snapshotting the volatile transformations error wrapper

arpl:
## 🔗 Ticket

No Linear ticket — unplanned CI repair. `test with code coverage` is currently red on `main` and on every open PR in the repo, so this is worth landing on its own. Happy to file one for traceability if you'd rather.

---

## Summary

`TestTransformationsTest/failure` snapshot-compares the upstream error reported for the deliberately broken Python library (`badPyLibrary`). The backend does not return a stable wrapper around the interpreter diagnostic, so pinning the exact string makes the test flap.

Both of these have come back from the same fixture, days apart, each passing CI at the time:

```
BadCodeError("'(' was never closed (<unknown>, line 1)")
SyntaxError(("Line 1: SyntaxError: '(' was never closed at statement: 'def broken_function('",))
```

Timeline on `main`:

| When | Snapshot expects | Backend returned | Result |
|---|---|---|---|
| #472 (fixtures added) | `BadCodeError` | `BadCodeError` | pass |
| Jul 23, #709 (`8b4a90ab`) | changed to `SyntaxError` | `SyntaxError` | pass |
| Jul 24, #680 (`7393796`) | `SyntaxError` | `BadCodeError` | **fail** |

So this is not one-directional drift that a re-baseline fixes — whichever form is pinned, the test goes red when the other is served. Re-baselining to `BadCodeError` today would just move the coin flip.

Worth flagging separately for the transformations owners: the flip to `SyntaxError` came in via [#709](https://github.com/rudderlabs/rudder-iac/pull/709) ("destination stability fixes and presence-based secrets"), whose only change under `cli/tests` was that one snapshot line. If the wrapper class alternating between `BadCodeError` and `SyntaxError` is *not* expected backend behaviour, the real bug is upstream and this PR only stops the CLI test suite from being the thing that reports it.

---

## Changes

- Exempt `libraries[].message` from the snapshot comparison in `testResultsIgnoreFields`, alongside the existing dynamic fields (`id`, `versionId`, `externalId`).
- Add `verifyBadLibraryMessage`, asserting directly that `badPyLibrary` reports `pass: false` and that its message contains the interpreter's own `'(' was never closed` diagnostic — the part common to both wrappers and the part the fixture actually exercises.
- Set the snapshot's `message` to `"placeholder"`, matching the convention already used for the other ignored fields.

Assertion strength is preserved: previously a wrong-but-well-formed error message would be caught only by exact match; now it is caught by the substring plus `pass: false`, while the wrapper class is free to vary.

---

## Testing

- `make lint` — 0 issues.
- `go vet ./cli/tests/...`, `go build ./...`, `go test ./cli/tests/helpers/...` — all clean.
- The e2e suite itself needs `RUDDERSTACK_ACCESS_TOKEN` against the shared CI workspace, so **I could not run `TestTransformationsTest` locally** — this PR's own CI run is the real verification. Note that a single green run does not by itself prove the flake is gone (a re-baseline would also go green if the matching form happened to be served); the argument is that the varying field is no longer compared at all.

---

## Risk / Impact

Low. Test-only change, no production code touched. The trade-off is deliberate: we stop asserting the exact upstream error string, which is the field demonstrated to be unstable.

---

## Checklist

- [ ] Ticket linked (none — see above)
- [x] Tests added/updated

### DIFF
--- a/cli/tests/command_transformations_test.go
+++ b/cli/tests/command_transformations_test.go
@@ -18,6 +18,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// badPyLibraryHandle is the handle of the intentionally broken Python library in the
+// failure fixture set.
+const badPyLibraryHandle = "badPyLibrary"
+
 func TestTransformationsTest(t *testing.T) {
 	executor, err := NewCmdExecutor("")
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


arpl:
Superseded by #712, which landed the same core fix (ignoring `libraries[].message` in `testResultsIgnoreFields`) about ten minutes ahead of this. Closing in favour of it — no point carrying two fixes for one red check.

Two small things from this PR that #712 doesn't cover, if anyone wants them as a follow-up:

- The snapshot still holds the stale `SyntaxError(...)` string at `libraries[0].message`. It's ignored now, so it's inert, but it will read as the expected value to the next person who opens the file. This PR set it to `"placeholder"`, matching the convention already used for the other ignored fields.
- Ignoring the message drops all assertion on the error *content*; `pass`/`status` alone would still be satisfied by a wrong-but-failing message. This PR kept a narrow check (`verifyBadLibraryMessage`) on `'(' was never closed`, the interpreter diagnostic common to both wrapper forms the backend returns.

Worth recording either way: this was not a one-directional upstream reword. Both wrappers have come back from the same fixture and each passed CI at the time — `BadCodeError` under #472, `SyntaxError` under #709 on Jul 23, then `BadCodeError` again on Jul 24. Ignoring the field is right precisely because it alternates; a re-baseline either way would have flapped.

PR #718, opened 2026-07-30
chore(cli): add S3 destination apply e2e test

Jayachand:
## 🔗 Ticket

resolves [DEX-565](https://linear.app/rudderstack/issue/DEX-565/add-s3-destination-apply-e2e-test-both-auth-modes)

---

## Summary

Adds an end-to-end test that applies the S3 destination through the CLI and verifies the upstream result, covering both auth modes (key-based access keys and role-based IAM ARN). The test drives destroy → apply create → apply update → re-apply (no-op) against a live stack, and both S3 variants are applied together from a single project folder — the same one-shot apply style the catalog e2e uses for events/properties. A new `DestinationSnapshotTester` helper mirrors the catalog `UpstreamSnapshotTester` so verification follows the established file-manager + count-guard + per-resource-compare convention.

---

## Changes

* Add `TestDestinationsApply` e2e test (`cli/tests/command_destinations_apply_test.go`).
* Add `DestinationSnapshotTester` helper (`cli/tests/helpers/snapshot_destinations.go`) — URN-keyed, count-guarded snapshot comparison mirroring `UpstreamSnapshotTester`.
* Add S3 destination specs (key-based + role-based, create/update + vars) under `cli/tests/testdata/project/destinations/s3/`.
* Add expected upstream snapshots under `cli/tests/testdata/expected/upstream/destinations/`.
* Update `.claude/linear-details.mdc` to the Destinations CLI project / DevEx team / Onboard Terraform Supported Destinations milestone.

---

## Testing

* `make test-e2e` — `TestDestinationsApply` passes against a live stack (create + update snapshots verified for both auth modes).
* `make lint` — 0 issues.
* `go vet ./tests/...` and `go test ./tests/helpers/...` pass.

---

## Risk / Impact

Low
* Test-only change plus one new test helper; no production code touched.
* The e2e requires the destination experimental flags and a live backend that has S3 (unverified) destination support enabled.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated


cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// A leftover unverified destination breaks other e2e tests' destroy when
+	// UnverifiedDestinations is off. Registered after t.Setenv so the flags still
+	// hold when this cleanup runs.
+	t.Cleanup(func() {
+		out, err := executor.Execute(cliBinPath, "destroy", "--confirm=false")
+		require.NoError(t, err, "cleanup destroy failed: %s", out)
+	})
+
+	apply := func(dir string) {
```
dhawal1248: The `apply` closure captures the parent test's `t` but is called from inside subtests. `require.NoError` on the parent `t` from a subtest goroutine attributes the failure to the parent test, and `FailNow`'s `runtime.Goexit` makes the subtest die with Go's cryptic `test executed panic(nil) or runtime.Goexit` instead of the real failure message.

Suggest passing the subtest's `t` explicitly, matching the catalog test's pattern:

```go
apply := func(t *testing.T, dir string) {
	t.Helper()
	...
}
```

and calling `apply(t, "create")` etc. inside each subtest.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,109 @@
...
+		apply("create")
+		verifyDestinationState(t, "create")
+	})
+
+	t.Run("apply update", func(t *testing.T) {
+		apply("update")
+		verifyDestinationState(t, "update")
+	})
+
+	t.Run("re-apply leaves upstream state unchanged", func(t *testing.T) {
```
dhawal1248: This subtest doesn't actually assert a no-op: it re-applies and snapshot-compares, but `version` and `updatedAt` are in the ignore list, so a spurious `Update` on re-apply would still pass — as written it's mostly a duplicate of the "apply update" subtest.

The catalog test covers this with `verifyNoChangesToApply`: `--dry-run` + assert the output contains "No changes to apply". Doing the same here would make the subtest verify what its name claims.

cli/tests/helpers/snapshot_destinations.go
```diff
@@ -0,0 +1,93 @@
...
+// DestinationLister is the subset of the destinations API client the tester needs.
+type DestinationLister interface {
+	GetAll(ctx context.Context) ([]client.Destination, error)
+}
+
+// DestinationSnapshotTester compares the CLI-managed destinations fetched from the
+// API against expected snapshot files, mirroring UpstreamSnapshotTester. Unlike
+// the catalog client, the destinations API exposes no hasExternalId list filter,
+// so managed destinations are selected client-side by ExternalID presence.
+type DestinationSnapshotTester struct {
```
dhawal1248, line 20: Missing unit tests for `DestinationSnapshotTester` — the sibling testers both have them (`snapshot_upstream_test.go`, `snapshot_transformations_test.go`) and CLAUDE.md makes unit tests mandatory. `DestinationLister` makes this trivially mockable; worth covering the ExternalID filtering, the count-mismatch error, and error aggregation.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	out, err := executor.Execute(cliBinPath, "destroy", "--confirm=false")
+	require.NoError(t, err, "destroy failed: %s", out)
+
+	// A leftover unverified destination breaks other e2e tests' destroy when
+	// UnverifiedDestinations is off. Registered after t.Setenv so the flags still
+	// hold when this cleanup runs.
+	t.Cleanup(func() {
+		out, err := executor.Execute(cliBinPath, "destroy", "--confirm=false")
+		require.NoError(t, err, "cleanup destroy failed: %s", out)
```
dhawal1248: nit: `require.NoError` inside `t.Cleanup` works (cleanups run on the test goroutine), but `t.Errorf`/`assert` is the safer convention in cleanups — `FailNow` mid-cleanup is easy to get wrong later.

cli/tests/command_destinations_apply_test.go
dhawal1248: Essentially what this means is - the test should assert that re-apply should be a no-op resulting in no operations from our side and hence no fields should change at all (including ignore fields)

dhawal1248 (changes requested):

cli/tests/command_destinations_apply_test.go
Jayachand: Done

cli/tests/command_destinations_apply_test.go
Jayachand: Good catch on the ignore-list masking. I looked into the `--dry-run` + "No changes to apply" approach, but it can't work for this scenario as written: the S3 key-based spec's `access_key_id`/`access_key` are write-only, so the API never returns them and the CLI maps them to **always-unknown** secrets (`secret.String.Diff` returns true unconditionally — see the "always-re-apply" note in `secret.go`). A dry-run would therefore always report a pending `Update` for the secret, and "No changes to apply" would never appear. This is the same reason the accounts e2e (`TestAccountsApply`) uses a snapshot check rather than a no-op assertion for its re-apply.

Since both S3 specs apply together, a folder-level dry-run always diffs. I've renamed the subtest to `re-apply churns only the write-only secret` and added a comment explaining the constraint; the snapshot of the non-secret upstream fields proves nothing else changes on re-apply. If we want a genuine no-op assertion, the clean way is to split the role-based (secret-free) spec into its own dry-run — happy to do that in a follow-up if you'd prefer.

cli/tests/helpers/snapshot_destinations.go
Jayachand: Added `cli/tests/helpers/snapshot_destinations_test.go` with a mock `DestinationLister`, covering the happy path, ExternalID filtering (unmanaged destinations excluded), the count-mismatch error, and per-URN error aggregation on a comparison mismatch.

cli/tests/command_destinations_apply_test.go
Jayachand: Switched to `assert.NoError` inside the cleanup.

cli/tests/command_destinations_apply_test.go
Jayachand: Agreed on the intent — a true no-op should change nothing at all. The blocker is the write-only secret: it's always-unknown on re-apply, so the plan always contains an Update for it and nothing we assert can make that a zero-op for the key-based spec (details in the thread above). The snapshot verifies the non-secret fields stay put. A secret-free (role-based) dry-run is the way to assert a real no-op — can add that separately if you'd like.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,115 @@
...
+// TestDestinationsApply drives the destination provider end-to-end against a live
+// stack: apply create → apply update → re-apply is a no-op. All destination specs
+// live side by side under one project folder and are applied together (like the
+// catalog e2e applies events, properties, etc. in one shot), then the managed
+// destinations are snapshot-compared upstream via DestinationSnapshotTester — the
+// same file-manager + count-guard + per-resource-compare structure verifyState
+// uses. Destinations are double-gated — DestinationSupport enables the kind,
+// UnverifiedDestinations registers the unverified ones (e.g. S3) — and key-auth
+// specs reference secrets via {{ .VAR }} placeholders resolved at apply time.
+func TestDestinationsApply(t *testing.T) {
```
sivashanmukh, line 57: **Gate this behind an env check, like the accounts e2e.**

This test runs on every `make test-e2e`, but the Risk section of the PR says it needs "a live backend that has S3 (unverified) destination support enabled." `TestAccountsApply` gates on exactly that kind of precondition:

```go
if os.Getenv("RUN_ACCOUNT_E2E") != "1" {
    t.Skip("set RUN_ACCOUNT_E2E=1 with a live stack (SOURCE_BIGQUERY definition + PAT) to run")
}
```

Why it matters more here than a plain red test: the unconditional `destroy` on line 49 fires *before* the apply that would fail. On a stack without unverified destinations, a dev running the suite loses everything in the workspace and only then sees the failure. The skip has to come before the destroy.

```suggestion
func TestDestinationsApply(t *testing.T) {
	if os.Getenv("RUN_DESTINATION_E2E") != "1" {
		t.Skip("set RUN_DESTINATION_E2E=1 with a live stack that has unverified destinations (S3) enabled")
	}

```

(needs `"os"` in the import block)

If the shared e2e stack is guaranteed to have `RUDDERSTACK_X_UNVERIFIED_DESTINATIONS` support, then drop the guard and say so in the doc comment instead — either is fine, but right now the precondition only exists in the PR description.

cli/tests/helpers/snapshot_destinations.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// The destinations API has no hasExternalId filter, so drop UI/other-tool
+	// resources here — only CLI-managed destinations carry an external ID.
+	managed := make(map[string]client.Destination)
+	for _, dest := range all {
+		if dest.ExternalID != "" {
+			managed[urn(destination.DestinationResourceType, dest.ExternalID)] = dest
+		}
+	}
```
sivashanmukh, line 55: **The count guard can be defeated by duplicate external IDs.**

`managed` is keyed by URN, so two upstream destinations sharing an `ExternalID` collapse into one entry. `len(managed)` then equals the expected file count and the guard passes, silently hiding one of them — the exact "unexpected create upstream" case the guard exists to catch. The new `count mismatch fails` unit case uses distinct IDs (`s3` / `s3-extra`), so nothing covers this today.

```suggestion
	// The destinations API has no hasExternalId filter, so drop UI/other-tool
	// resources here — only CLI-managed destinations carry an external ID.
	managed := make(map[string]client.Destination)
	for _, dest := range all {
		if dest.ExternalID == "" {
			continue
		}

		resourceURN := urn(destination.DestinationResourceType, dest.ExternalID)
		if _, dup := managed[resourceURN]; dup {
			return fmt.Errorf("duplicate managed destination for %s", resourceURN)
		}
		managed[resourceURN] = dest
	}
```

Worth a fifth unit case now that the test file exists — two destinations both with `ExternalID: "s3"`, asserting the error.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/rudderlabs/rudder-iac/cli/tests/helpers"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// destinationSnapshotIgnore are the volatile upstream fields excluded from the
+// snapshot comparison: server-assigned id, workspace scoping, version, and
+// timestamps. Write-only secrets (e.g. S3 access keys) are never returned by the
+// API, so they never appear here.
+var destinationSnapshotIgnore = []string{"id", "workspaceId", "version", "createdAt", "updatedAt"}
```
sivashanmukh, line 24: **Please note the assumption this list depends on.**

The ignore list only suppresses the *value* comparison. `compareMaps` reports `missing key '<k>' in actual` before `compareValues` ever consults `ignore`, so a field being absent still fails even when it's listed here.

That matters because `client.Destination` marks all four of these `omitempty` (`Version int64`, `CreatedAt/UpdatedAt *time.Time`, `WorkspaceID string`). If the API ever returns `version: 0` or null timestamps, these four expected files fail on the very fields you meant to ignore, with a confusing "missing key" message.

Two lines of comment naming that would save the next person the bisect — something like: "listed fields are ignored by value, not by presence: the API must still return each key or CompareStates reports a missing key."

(Setting these fields in the `s3Destination()` unit mock + its fixture would also turn this into something the unit test can catch without a live stack.)

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,115 @@
...
+	t.Setenv("RUDDERSTACK_X_ENABLE_VAR_SUBSTITUTION", "true")
+
+	executor, err := NewCmdExecutor("")
+	require.NoError(t, err)
+
+	projectDir := filepath.Join("testdata", "project", "destinations", "s3")
+	varFile := filepath.Join(projectDir, "s3.vars.yaml")
+
+	out, err := executor.Execute(cliBinPath, "destroy", "--confirm=false")
+	require.NoError(t, err, "destroy failed: %s", out)
```
sivashanmukh, line 74: Nit: `destroy` renders a plan of what it's about to delete, so its output is a leak surface too — but only the `apply` output is scanned against `destinationRawSecrets`. Cheap to fold in, e.g. hoist the loop into a small `assertNoRawSecrets(t, out)` helper and call it after both `destroy` invocations as well.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+// destinationRawSecrets are literal secret values from the var file that must
+// never surface in CLI output.
+var destinationRawSecrets = []string{
+	"AKIAIOSFODNN7EXAMPLE",
+	"wJalrXUtnFEMI-K7MDENG-bPxRfiCYEXAMPLEKEY",
+}
+
+// TestDestinationsApply drives the destination provider end-to-end against a live
+// stack: apply create → apply update → re-apply is a no-op. All destination specs
```
sivashanmukh: Stale after the last commit: this still says "re-apply is a no-op", which now contradicts both the renamed subtest (`re-apply churns only the write-only secret`) and the comment you added above it explaining that a dry-run would always report a diff. Suggest "apply create -> apply update -> re-apply churns only the write-only secret".

sivashanmukh (commented):
Re-review of the updated branch. The helper unit test, the `require`->`assert` in cleanup, and the `t.Helper()` fix all look good, and the re-apply justification is sound - I've dropped that finding. Five things left, inline. Two of them (the env gate and the count guard) are worth acting on before merge; the rest are comment/nit level.

Separately, and not blocking: the fixtures still live under `testdata/project/`, which is `TestProjectApply`'s root, while the accounts e2e keeps its own at `testdata/accounts/`. Nothing breaks today since that test only targets `project/create` and `project/update`, but `git mv cli/tests/testdata/project/destinations cli/tests/testdata/destinations` plus one line in this file makes the ownership obvious.

sivashanmukh (commented):
Two follow-ups from the earlier review, now as concrete suggestions: the fixture relocation, and enriching the unit mock so it matches the shape the API actually returns. I applied the second one locally to check - the suite passes with the fields set, and it correctly fails with `missing key 'version' in actual` if `omitempty` drops one, which is exactly the ignore-list blind spot I raised.

cli/tests/command_destinations_apply_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestDestinationsApply(t *testing.T) {
+	t.Setenv("RUDDERSTACK_X_DESTINATION_SUPPORT", "true")
+	t.Setenv("RUDDERSTACK_X_UNVERIFIED_DESTINATIONS", "true")
+	t.Setenv("RUDDERSTACK_CLI_EXPERIMENTAL", "true")
+	t.Setenv("RUDDERSTACK_X_ENABLE_VAR_SUBSTITUTION", "true")
+
+	executor, err := NewCmdExecutor("")
+	require.NoError(t, err)
+
+	projectDir := filepath.Join("testdata", "project", "destinations", "s3")
```
sivashanmukh: **Move the fixtures out of `testdata/project/`.**

`testdata/project` is `TestProjectApply`'s root, and the accounts e2e keeps its own fixtures at `testdata/accounts/`. These specs plus a stray `s3.vars.yaml` now live inside another test's project tree.

Nothing breaks today, because `TestProjectApply` only ever targets `testdata/project/create` and `testdata/project/update` (and `copyAndMigrateProject` copies just those two). But it's a latent trap: anything that later points `-l`, `validate`, or `import` at `testdata/project` recursively picks up destination specs it wasn't written for, and the failure mode would be confusing.

One `git mv`:

```bash
git mv cli/tests/testdata/project/destinations cli/tests/testdata/destinations
```

and one line here:

```suggestion
	projectDir := filepath.Join("testdata", "destinations", "s3")
```

`testdata/expected/upstream/destinations/` can stay where it is - it already sits next to `expected/upstream/accounts/`.

cli/tests/helpers/snapshot_destinations_test.go
```diff
@@ -0,0 +1,86 @@
...
+		Name:       "My S3",
+		Type:       "S3",
+		IsEnabled:  true,
+		Config: json.RawMessage(`{
+			"bucketName": "my-bucket",
+			"roleBasedAuth": true,
+			"iamRoleARN": "arn:aws:iam::123456789012:role/S3Access"
+		}`),
+	}
+}
```
sivashanmukh: **The comment says `id/workspaceId`, but only `ID` is set** - and the fixture omits `workspaceId`, `version`, `createdAt`, `updatedAt` entirely, so the ignore list is only exercised for `id`.

Worth setting all of them, because the live API *does* return those four (that's why `destinationSnapshotIgnore` names them and all four e2e expected files carry them). As it stands, the unit fixture has a different shape from the thing it's standing in for.

Concretely, it also makes the presence-vs-value gap I flagged on `destinationSnapshotIgnore` catchable without a live stack. I tried it locally: with the fields set the suite passes, and flipping `Version` to `0` so `omitempty` drops the key fails with

```
mismatch at path '': missing key 'version' in actual
```

even though `version` is in the ignore list. That's the regression the e2e would otherwise hit blind against a real backend.

```suggestion
// s3Destination mirrors the destination_s3 snapshot fixture, including the
// volatile fields the API always returns (id/workspaceId/version/timestamps)
// that the ignore list must drop.
func s3Destination() client.Destination {
	var (
		created = time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
		updated = time.Date(2026, 1, 2, 0, 0, 0, 0, time.UTC)
	)

	return client.Destination{
		ID:          "srv-generated-id",
		WorkspaceID: "ws-123",
		Version:     3,
		CreatedAt:   &created,
		UpdatedAt:   &updated,
		ExternalID:  "s3",
		Name:        "My S3",
		Type:        "S3",
		IsEnabled:   true,
		Config: json.RawMessage(`{
			"bucketName": "my-bucket",
			"roleBasedAuth": true,
			"iamRoleARN": "arn:aws:iam::123456789012:role/S3Access"
		}`),
	}
}
```

(needs `"time"` in the import block, and the four keys added to the fixture - see the next comment)

cli/tests/helpers/testdata/snapshot/destinations/destination_s3
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "<ignored>",
```
sivashanmukh, line 2: Companion to the `s3Destination()` suggestion - `CompareStates` compares key counts, so the fixture needs the same four keys:

```suggestion
  "id": "<ignored>",
  "workspaceId": "<ignored>",
  "version": "<ignored>",
  "createdAt": "<ignored>",
  "updatedAt": "<ignored>",
```

The `"<ignored>"` string against a numeric `version` / RFC3339 timestamp is fine: `compareValues` checks the ignore path before the type check, which is the same trick the four e2e expected files already use.

cli/tests/command_destinations_apply_test.go
Jayachand: Added the gate: `TestDestinationsApply` now skips unless `RUN_DESTINATION_E2E=1`, placed before the destroy so a stack without unverified-destination support isn't wiped before the failing apply. Also note I'll want this to eventually cover both verified and unverified destinations, so the gate keys off a dedicated run flag rather than hard-coding the unverified assumption.

cli/tests/helpers/snapshot_destinations.go
Jayachand: I'm going to leave this one as-is. ExternalID uniqueness is a backend invariant (DB constraint) — two remote destinations can't share an ExternalID — and the whole destination provider already relies on it: `MapRemoteToState` and `LoadRemoteResources` key state by ExternalID too, so a duplicate would collapse there long before this test guard, and if the invariant were ever violated the provider itself would break first, not this snapshot check. Adding duplicate detection here would be guarding an input the backend forbids and the provider already assumes away (CLAUDE.md: no error handling for impossible scenarios). Happy to reconsider if there's a path where GetAll can actually return duplicate ExternalIDs.

cli/tests/command_destinations_apply_test.go
Jayachand: Documented — added a note on the ignore list that these fields are ignored by value, not by presence (CompareStates reports a missing key before consulting the ignore list, so the API must still return each key). Also set all four volatile fields in the unit fixture so this is now caught without a live stack (see the s3Destination() / fixture changes).

cli/tests/command_destinations_apply_test.go
Jayachand: Done — hoisted the scan into assertNoRawSecrets(t, out) and call it after both the up-front destroy and the cleanup destroy, not just apply.

cli/tests/command_destinations_apply_test.go
Jayachand: Fixed — updated the doc comment to 'apply create -> apply update -> re-apply churns only the write-only secret', matching the renamed subtest.

cli/tests/command_destinations_apply_test.go
Jayachand: Moved — `git mv cli/tests/testdata/project/destinations cli/tests/testdata/destinations` and updated projectDir to `testdata/destinations/s3`. Expected snapshots stay under `testdata/expected/upstream/destinations/` next to accounts, as you noted.

cli/tests/helpers/snapshot_destinations_test.go
Jayachand: Applied — s3Destination() now sets id/workspaceId/version/createdAt/updatedAt. I verified your regression locally: flipping Version to 0 (so omitempty drops the key) fails with `missing key 'version' in actual` even though version is in the ignore list, and setting it back passes. So the presence-vs-value gap is now caught by the unit test without a live stack.

cli/tests/helpers/testdata/snapshot/destinations/destination_s3
Jayachand: Done — fixture now carries all five ignored keys (id/workspaceId/version/createdAt/updatedAt) as "<ignored>", matching the marshaled shape so the key-count check passes.

dhawal1248 (approved):

PR #721, opened 2026-07-30
test(cli): e2e for accounts import workspace (scaffold, adopt, on-disk secret-leak grep)

sivashanmukh:
Closes the E2E half of [DEX-461](https://linear.app/rudderstack/issue/DEX-461) — tracked as [DEX-569](https://linear.app/rudderstack/issue/DEX-569).

The accounts e2e that shipped with #710 (`TestAccountsApply`) covers only the **apply** path. DEX-461 also asked for the import leg, a managed-check, and a secret-leak assertion on generated files. This adds them.

## Gap this closes

| DEX-461 requirement | before | now |
| -- | -- | -- |
| fill var → `apply` → re-`apply` no-op | ✅ | ✅ |
| modify options → `apply` updates | ✅ | ✅ |
| `import workspace` scaffolds specs + var file with `{{ .VAR }}` | ❌ absent | ✅ |
| account is managed afterwards (no longer importable) | ⚠️ only `hasExternalID=true` | ✅ asserts it left the importable set |
| secrets never in state or exported specs | ⚠️ CLI stdout only | ✅ walks every generated file |

## What the test does

`cli/tests/command_accounts_import_test.go`, same `RUN_ACCOUNT_E2E=1` gate:

1. `destroy` — `import workspace` refuses to run while the project has changes to sync, so the managed set must be empty first (same precondition `TestAccountsApply` establishes).
2. Seeds an **unmanaged** `SOURCE_BIGQUERY` account via the API client (no `externalId`) carrying the raw credential; deleted on cleanup.
3. `import workspace` into a temp dir → finds the scaffolded spec **by name**, asserts `credentials` came out as a `{{ .VAR }}` reference, and that `imported/secrets.vars.yaml` holds the matching placeholder.
4. **Walks every generated file** and fails on the raw secret — the assertion that matters, since `FormatForExport` → `secret.MaskSecrets` is what keeps a plaintext credential out of version control.
5. Prunes the generated project to the seeded account — both the specs and the import manifest, since a manifest entry whose spec is gone is an orphaned URN.
6. Fills the placeholder → `apply` adopts the account (import only scaffolds; the remote is claimed on apply).
7. Asserts adoption: the same upstream id now carries an `externalId`, has left the importable set, and never returns the credential in `options`.
8. Re-`apply` stays clean.

## Testing

**Run against a live workspace** (`RUN_ACCOUNT_E2E=1 go test ./cli/tests/ -run TestAccountsImportWorkspace`). The first run failed, for two reasons worth recording — both were flaws in the test, not the provider:

- `import workspace` scaffolds **every** importable resource in the workspace, so the apply leg reconciled the whole workspace and died on an unrelated transformation that does not validate upstream (`TypeError: Cannot set properties of undefined`). Fixed by pruning to the seeded account before applying.
- Requiring exactly one scaffolded account spec assumed the workspace held no other importable accounts. Fixed by matching on name.

The import + masking + var-file assertions (steps 3–4, the substance of this PR) passed on that run before the apply leg was reached.

The corrected version has not been re-run end to end: doing so requires `destroy` against the target workspace, and the only stack available here was a personal production workspace. **This suite needs a disposable workspace** — that requirement is now called out in the test's doc comment. `go vet ./cli/tests/` is clean and the test compiles and skips cleanly without the gate.


cli/tests/command_accounts_import_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	t.Fatalf("no scaffolded spec found for account %q among %v", importedAccountName, matches)
+	return "", ""
+}
+
+// pruneToAccount reduces the generated project to the seeded account: every
+// other scaffolded spec is removed, and the import manifest is filtered to the
+// matching URN. Both halves are needed — a manifest entry whose spec is gone is
+// an orphaned URN and fails validation.
+func pruneToAccount(t *testing.T, importedDir, keepSpec, remoteID string) {
```
Jayachand, line 202: **The spec and the manifest are pruned by two different identities — nothing ties them together.**

The kept spec is chosen by *name match* (`findAccountSpec` greps for `importedAccountName`), while the manifest is filtered by *remote ID*. If the workspace ever holds another importable account whose spec content happens to contain the same name string, the prune can keep a spec and a manifest entry that refer to different accounts — and the failure surfaces as a confusing apply/validation error rather than a clear test failure.

Cheap hardening: after filtering, assert the kept manifest entry's URN appears in the kept spec's content, so a mismatch fails loudly at the prune step.

cli/tests/command_accounts_import_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	err := filepath.WalkDir(dir, func(path string, entry os.DirEntry, err error) error {
+		if err != nil || entry.IsDir() {
+			return err
+		}
+		content, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+		assert.NotContains(t, string(content), rawAccountSecret,
```
Jayachand, line 279: **`assertNoSecretOnDisk` greps for only `rawAccountSecret`, while its sibling `assertNoSecretsInOutput` checks all four fixture secrets.**

Correct today, since this test only seeds a BigQuery account — but the two helpers holding divergent notions of "the secrets" is a trap the moment someone seeds a PG/SF account here. Since `fixtureAccountSecrets` already exists, loop over it:

```go
for _, s := range fixtureAccountSecrets {
    assert.NotContains(t, string(content), s, "raw secret leaked into %s", ...)
}
```

(Longer-term, passing the secret slice as a parameter would make this helper reusable by the catalog/destination import e2es that will follow once importMerge ships.)

cli/tests/command_accounts_import_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	out, err = executor.Execute(cliBinPath, "apply", "-l", projectDir, "--var-file", varFile, "--confirm=false")
+	require.NoError(t, err, "apply after import failed: %s", out)
+	assertNoSecretsInOutput(t, out)
+
+	// Adopted rather than recreated: the same upstream account now carries an
+	// external id, so it has left the importable set.
+	adopted, err := apiClient.Accounts.Get(ctx, seededID)
+	require.NoError(t, err, "the seeded account must still exist after apply")
+	assert.NotEmpty(t, adopted.ExternalID, "apply must claim the account with an external id")
```
Jayachand, line 118: **The adoption check is weaker than it reads — consider the suite's snapshot approach instead.**

`NotEmpty(adopted.ExternalID)` passes even if apply claimed the account under the *wrong* external ID, and nothing checks name/options survived adoption intact. The sibling `TestAccountsApply` already does this properly: JSON round-trip → `helpers.CompareStates` against `testdata/expected/upstream/accounts/<dir>/<externalId>.json` with the volatile fields ignored. A `.../imported/<id>.json` snapshot here would assert the whole adopted account (per the repo convention of whole-struct over field-by-field), and would have this test verifying adoption as strongly as the apply test verifies create/update.

The `NotContains(options, rawAccountSecret)` line should stay regardless — a snapshot can't express that negative.

cli/tests/command_accounts_import_test.go
```diff
@@ -0,0 +1,284 @@
...
+	assert.Contains(t, spec, "SOURCE_BIGQUERY", "scaffolded spec must carry the account definition")
+
+	reference := varReference.FindStringSubmatch(spec)
+	require.NotNil(t, reference, "credentials must be exported as a {{ .VAR }} reference, got:\n%s", spec)
+	varName := reference[1]
+
+	varFile := filepath.Join(importedDir, importer.SecretsVarFileName)
+	scaffoldedVars, err := os.ReadFile(varFile)
+	require.NoError(t, err, "import must scaffold %s", importer.SecretsVarFileName)
+	assert.Contains(t, string(scaffoldedVars), varName+":",
```
Jayachand, line 96: **Substring match can pass on the wrong key.**

Since the var file is a flat `KEY:` list, `Contains(varName+":")` also matches a *different, longer* key that merely ends with `varName` — e.g. a captured `SECRET` would match a `MY_SECRET:` line. Anchoring to line start makes it exact:

```go
assert.Regexp(t, "(?m)^"+regexp.QuoteMeta(varName)+":", string(scaffoldedVars), ...)
```

Jayachand:
Given Import e2e tests are not adopted earlier and had to be implemented for other resources such as catalog, destinations as well. Would like to take some time on how to establish import e2e framework for all.
Also X_IMPORT_MERGE feature is merged, which also affects framework once enabled

PR #726, opened 2026-08-05
fix(destination): prune empty non-required config fields on import export

Jayachand:
## 🔗 Ticket

resolves [DEX-612](https://linear.app/rudderstack/issue/DEX-612/prune-empty-non-required-config-fields-on-destination-import)

---

## Summary

Importing a destination produced cluttered, misleading YAML: the backend echoes the full config schema with every field populated (empty strings, empty arrays, arrays of empty objects), the converter copies every present key, and `MaskSecrets` then wraps empty secret fields into `{{ .VAR }}` references. For a noAuth HTTP destination this meant phantom `password`/`bearer_token`/`api_key_value` placeholders plus a pile of empty fields. This change prunes a config field on the import/export path when it is **empty AND not required** given the config's other values, using the same validator project validation uses (so `required_if` conditionals are honored). The apply/converter path is untouched.

---

## Changes

* Add `RegisteredDefinition.PruneEmptyOptional` (`definitions/prune.go`): drops empty, non-required fields; keeps required-if-conditional empties (e.g. basicAuth username/password), non-empty optionals, and bools.
* Wire the prune into `toExportSpecMap` before `MaskSecrets`, so absent secrets never become masked `{{ .VAR }}` placeholders.
* Unit tests: prune logic + empty-value detection, http auth-mode matrix (noAuth drops / basicAuth keeps / batching keeps max_batch_size / user-set optional survives), and an end-to-end `FormatForExport` case.

---

## Testing

* Unit: `go test ./cli/internal/providers/destination/...` — all pass.
* Build: `go build ./...`; Lint: `make lint` (0 issues).
* Manual: imported a live noAuth HTTP destination → clean minimal spec (no phantom secrets/empties); `apply` then re-`apply` of the imported spec is idempotent ("No changes to apply"). Confirmed pruning is export-only — not in Create/Update/LocalToAPI.

---

## Risk / Impact

Low — import/export output only; the apply cycle and converter are unchanged, so no phantom-diff risk. Only affects the human-facing imported YAML.

---

## Checklist

* [x] Ticket linked
* [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/destination/definitions/http/prune_test.go
+++ b/cli/internal/providers/destination/definitions/http/prune_test.go
@@ -1,0 +1,141 @@
+package http_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/rudderlabs/rudder-iac/cli/internal/providers/destination/definitions"
+	httpdest "github.com/rudderlabs/rudder-iac/cli/internal/providers/destination/definitions/http"
+)
+
+func registeredHTTP(t *testing.T) *definitions.RegisteredDefinition {
+	t.Helper()
+	registry := definitions.NewRegistry()
+	require.NoError(t, registry.Register(httpdest.NewDefinition()))
+	registered, err := registry.Get("http", 1)
+	require.NoError(t, err)
+	return registered
+}
+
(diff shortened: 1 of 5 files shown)


Jayachand:
duplicate

PR #727, opened 2026-08-05
feat(destination): onboard google pub/sub destination definition

Jayachand:
## 🔗 Ticket

Resolves [DEX-505](https://linear.app/rudderstack/issue/DEX-505/onboard-destination-google-pubsub)

---

## Summary

Onboards Google Cloud Pub/Sub into the CLI destination registry, following the `onboard-cli-destination` skill. Property mappings are ported from terraform-provider-rudderstack, validations derived from the integrations-config `schema.json`, and source types / connection modes / secret keys from `db-config.json`.

Note on naming: the ticket refers to `google_pubsub` (terraform's registration name), but the integrations-config directory is `googlepubsub` and its `db-config.json` declares `name: "GOOGLEPUBSUB"`. The canonical CLI type is therefore **`googlepubsub`**.

Example spec:

```yaml
version: rudder/v1
kind: destination
metadata:
  name: my-googlepubsub-destination
spec:
  id: my-googlepubsub-destination
  display_name: My Google Cloud PubSub Destination
  type: googlepubsub
  enabled: true
  definition_version: 1
  config:
    project_id: my-gcp-project
    credentials: "{{ .GOOGLE_PUBSUB_CREDENTIALS }}"
    event_to_topic_map:
      - from: Product Viewed
        to: product-events
      - from: Order Completed
        to: order-events
    event_to_attribute_map:
      - from: userId
        to: user_id
```

Usage requires `experimental: true` plus both `flags.destinationSupport` and `flags.unverifiedDestinations`.

---

## Changes

- Add `definitions/googlepubsub/definition.go` — `projectId`, `credentials` (secret), and the `eventToTopicMap` / `eventToAttributesMap` array-of-object reshapes, plus consent management via `common.Properties`.
- Register `googlepubsub` in `newDestinationRegistry` inside the `UnverifiedDestinations` gate; update the flag-matrix expected types.
- Add `definition_test.go` covering metadata, config validation (required fields, length patterns, unknown keys, consent), and conversion round-trips.
- Add destination e2e fixtures for two variations (topic-map only, and topic + attribute maps) with create/update pairs, matching upstream snapshots, and a dummy `GOOGLE_PUBSUB_CREDENTIALS` var plus its raw-secret leak assertion.

---

## Testing

- **Unit tests**: `go test ./cli/internal/providers/destination/... ./cli/internal/app/...` — all pass. Pattern validations were verified to be load-bearing by reverting the `validate` tags and confirming exactly the three pattern subtests fail, then restoring.
- **Conversion**: expected e2e snapshots were derived from actual `LocalToAPI` output rather than hand-written, confirming the `event_to_attribute_map` → `eventToAttributesMap` singular/plural mapping.
- **Manual**: every fixture and the example spec were validated through the real binary (`rudder-cli validate -l <dir> --var-file ...`), which caught two issues fixed in this PR:
  - `display_name` rejects `/`, so "Google Pub/Sub" fails the spec-syntax rule — renamed to "Google PubSub" in all fixtures and snapshots.
  - Substituting raw service-account JSON into `credentials` breaks YAML parsing, since the `{...}` value is spliced in unquoted. The e2e var file therefore uses an opaque placeholder. Worth noting for users, as service-account keys are natively JSON.
- **E2E**: `go test ./cli/tests -run TestDestinationsApply` and `./cli/tests/helpers -run TestDestinationSnapshotTester` compile and skip cleanly. The gated live run (`RUN_DESTINATION_E2E=1`) was **not** performed — no destination-enabled stack available — so snapshots are converter-derived and need one live run to confirm against real API responses.
- `make lint` — 0 issues. Rule-doc catalog is drift-free.

---

## Risk / Impact

Low

Notes: additive only — a new definition behind two experimental flags, with no changes to shared destination machinery. Existing destinations are untouched.

**One open question for review:** upstream `supportedSourceTypes` includes `amp`, `shopify`, and `warehouse`; this PR supports the same 10 types as the existing s3/rs definitions and drops those three. `amp` and `shopify` are clearly not CLI-owned, but `warehouse` is different — `common.SourceTypeWarehouse` already exists in the CLI's source-type mapping table, and unlike s3/rs, Pub/Sub genuinely lists `warehouse` upstream in both `supportedSourceTypes` and `supportedConnectionModes`. As written, a `consent_management.warehouse` block (valid per `schema.json`, which defines consent for all 13 types) is rejected. If warehouse sources should be able to feed Pub/Sub, adding it is a three-line change and consent coverage follows automatically. Flagging rather than deciding unilaterally, since it is a product question.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated


Jayachand:
duplicate

PR #738, opened 2026-08-17
feat: add externalId support to the connections API client

dhawal1248:
## 🔗 Ticket

Resolves [DEX-648](https://linear.app/rudderstack/issue/DEX-648/2-add-externalid-support-to-the-connections-api-client)

---

## Summary

Adds externalId support to the generic connections API client so event stream connections can carry IaC identity, mirroring the existing accounts (#638) and destinations externalId patterns in the same package.

---

## Changes

- `ExternalID` field on the `Connection` model (`externalId,omitempty`): read from API responses (tolerates the key being absent), and forwarded on `Create` so an externalId can be claimed atomically at create time
- `Update` strips `ExternalID` from the request body — externalId can only be set through the dedicated endpoint; the generic `PUT /v2/connections/{id}` rejects bodies carrying it
- `List` accepts variadic functional options with `WithConnectionsHasExternalID(bool)`, adding the `hasExternalId` query param used for CLI state loading (backward-compatible for existing callers)
- `SetExternalID(ctx, id, externalID)` calling `PUT /v2/connections/{id}/external-id`, mapping HTTP 409 to the new `ErrExternalIDAlreadyInUse` sentinel so callers can distinguish "already in use by another connection"

---

## Testing

- Unit tests for all new behaviour: externalId read on `Get`/`List`, forwarded on `Create`, stripped on `Update`, `hasExternalId=true|false` query encoding, `SetExternalID` request path/body, and 409 → `ErrExternalIDAlreadyInUse`
- `make lint` (0 issues) and full `make test` (0 failures)

---

## Risk / Impact

Low
Notes: purely additive client surface; existing callers are unaffected (`List` change is variadic, whole repo compiles unchanged).

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/api/client/connections.go
+++ b/api/client/connections.go
@@ -1,12 +1,23 @@
 package client
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/url"
+	"strconv"
 	"time"
 )
 
+// ErrExternalIDAlreadyInUse signals a 409 from the set external ID endpoint:
+// the external ID is already claimed by another connection in the workspace.
+var ErrExternalIDAlreadyInUse = fmt.Errorf("external ID already in use")
+
 type Connection struct {
 	ID            string     `json:"id,omitempty"`
+	ExternalID    string     `json:"externalId,omitempty"`
 	SourceID      string     `json:"sourceId"`
 	DestinationID string     `json:"destinationId"`
 	IsEnabled     bool       `json:"enabled"`
@@ -23,6 +34,18 @@ type ConnectionsPage struct {
 	Connections []Connection `json:"connections"`
 }
 
+type ListConnectionsOption func(*ListConnectionsOptions)
+
+func WithConnectionsHasExternalID(hasExternalID bool) ListConnectionsOption {
+	return func(o *ListConnectionsOptions) {
+		o.HasExternalID = &hasExternalID
+	}
+}
+
+type ListConnectionsOptions struct {
+	HasExternalID *bool
+}
+
 func (s *connections) Next(ctx context.Context, paging Paging) (*ConnectionsPage, error) {
 	page := &ConnectionsPage{}
 	ok, err := s.service.next(ctx, paging, page)
@@ -32,13 +55,32 @@ func (s *connections) Next(ctx context.Context, paging Paging) (*ConnectionsPage
 	return page, err
 }
 
-func (s *connections) List(ctx context.Context) (*ConnectionsPage, error) {
+func (s *connections) List(ctx context.Context, opts ...ListConnectionsOption) (*ConnectionsPage, error) {
 	page := &ConnectionsPage{}
-	if err := s.list(ctx, page); err != nil {
+	if err := s.list(ctx, page, opts...); err != nil {
 		return nil, err
 	}
 
 	return page, nil
+}
+
+func (s *connections) list(ctx context.Context, result interface{}, opts ...ListConnectionsOption) error {
+	options := &ListConnectionsOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	path := s.basePath
+	query := url.Values{}
+	if options.HasExternalID != nil {
+		query.Add("hasExternalId", strconv.FormatBool(*options.HasExternalID))
+	}
+	if len(query) > 0 {
+		path = fmt.Sprintf("%s?%s", path, query.Encode())
+	}
+
+	_, err := s.next(ctx, Paging{Next: path}, result)
+	return err
 }
 
 func (s *connections) Get(ctx context.Context, id string) (*Connection, error) {
@@ -64,9 +106,12 @@ func (s *connections) Create(ctx context.Context, connection *Connection) (*Conn
 }
 
 func (s *connections) Update(ctx context.Context, connection *Connection) (*Connection, error) {
-	// copy input and remove ID from request body without modifying input
+	// copy input and remove fields that should not be in request body without
+	// modifying input; externalId can only be set through the external-id
+	// endpoint, the generic update rejects bodies carrying it
 	conn := *connection
 	conn.ID = ""
+	conn.ExternalID = ""
 
 	response := struct{ Connection *Connection }{}
 	if err := s.update(ctx, connection.ID, &conn, &response); err != nil {
@@ -78,4 +123,22 @@ func (s *connections) Update(ctx context.Context, connection *Connection) (*Conn
 
 func (s *connections) Delete(ctx context.Context, id string) error {
 	return s.service.delete(ctx, id)
+}
+
+func (s *connections) SetExternalID(ctx context.Context, id, externalID string) error {
+	body, err := json.Marshal(map[string]string{"externalId": externalID})
+	if err != nil {
+		return fmt.Errorf("marshalling external ID: %w", err)
+	}
+
+	path := fmt.Sprintf("%s/%s/external-id", s.basePath, id)
+	if _, err := s.client.Do(ctx, "PUT", path, bytes.NewReader(body)); err != nil {
+		var apiError *APIError
+		if errors.As(err, &apiError) && apiError.HTTPStatusCode == 409 {
+			return ErrExternalIDAlreadyInUse
+		}
+		return fmt.Errorf("setting external ID: %w", err)
+	}
+
+	return nil
 }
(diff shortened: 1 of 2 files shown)


api/client/connections.go
```diff
@@ -1,12 +1,23 @@
...
+	"errors"
+	"fmt"
+	"net/url"
+	"strconv"
 	"time"
 )
 
+// ErrExternalIDAlreadyInUse signals a 409 from the set external ID endpoint:
+// the external ID is already claimed by another connection in the workspace.
+var ErrExternalIDAlreadyInUse = fmt.Errorf("external ID already in use")
```
dhawal1248:
```suggestion
var ErrExternalIDAlreadyInUse = fmt.Errorf("external ID already in use by another connection in the workspace")
```

api/client/connections_test.go
```diff
@@ -171,6 +173,49 @@ func TestClientConnectionsCreate(t *testing.T) {
...
+		ExternalID:    "external-id-1",
+		SourceID:      "source-id",
+		DestinationID: "destination-id",
+	})
+	require.NoError(t, err)
+	assert.NotNil(t, connection)
+	assert.Equal(t, "some-id", connection.ID)
+	assert.Equal(t, "external-id-1", connection.ExternalID)
+
+	httpClient.AssertNumberOfCalls()
```
dhawal1248: what does this do?

api/client/connections.go
sivashanmukh: **The doc no longer matches what a 409 means.**

`rudderlabs/rudder-config-backend#6576` makes connection `externalId` **set-once**, so `PUT /connections/:id/external-id` now returns 409 for two distinct conditions:

- the externalId is already in use by *another* connection (`A connection with this externalId already exists in the workspace`)
- *this* connection already has an externalId (`externalId is already set for this connection and cannot be changed`) — even when the value is identical

Only the message text separates them; the status is the same for both. Import (#754) is the first caller that can hit either, and it will report both as "claimed by another connection".

```suggestion
// ErrExternalIDAlreadyInUse signals a 409 from the set external ID endpoint.
// The backend uses 409 for two distinct conditions and distinguishes them only
// by message: the external ID is already claimed by another connection, or this
// connection already has one (externalId is set-once and cannot be changed).
// Callers that need to tell them apart must inspect the wrapped error.
var ErrExternalIDAlreadyInUse = fmt.Errorf("external ID already in use")
```

I have asked on the backend PR for a machine-readable `code` on the two conflict bodies so this can be branched on properly rather than by string matching. Until that lands, the wrapped error is the only signal — see the companion comment on the 409 branch below.

api/client/connections.go
```diff
@@ -79,3 +124,21 @@ func (s *connections) Update(ctx context.Context, connection *Connection) (*Conn
...
+	body, err := json.Marshal(map[string]string{"externalId": externalID})
+	if err != nil {
+		return fmt.Errorf("marshalling external ID: %w", err)
+	}
+
+	path := fmt.Sprintf("%s/%s/external-id", s.basePath, id)
+	if _, err := s.client.Do(ctx, "PUT", path, bytes.NewReader(body)); err != nil {
+		var apiError *APIError
+		if errors.As(err, &apiError) && apiError.HTTPStatusCode == 409 {
+			return ErrExternalIDAlreadyInUse
```
sivashanmukh: **Two things here.**

**1. The server message is discarded.** Returning the bare sentinel throws away the only thing that distinguishes "already claimed by another connection" from "this connection already has one" (see the comment on the sentinel declaration). Keeping both costs one line:

```suggestion
		if errors.As(err, &apiError) && apiError.HTTPStatusCode == 409 {
			return fmt.Errorf("%w: %w", ErrExternalIDAlreadyInUse, err)
		}
```

`errors.Is(err, ErrExternalIDAlreadyInUse)` still works for existing callers, and the backend's message survives for the user and for any caller that needs to tell the two cases apart.

**2. `Create` has no 409 mapping at all.** rudderlabs/rudder-config-backend#6576 also makes `POST /connections` return 409 when the body carries an `externalId` and a live connection already exists for the source/destination pair. #742's handler wraps that as `creating event stream connection %q: %w`, so the user sees the backend's raw advice to use the external-id endpoint — which is the wrong move for the CLI, since the existing connection typically has no externalId and wants adoption, not a set.

Worth a sibling sentinel so #754's import can offer adoption instead of surfacing a dead end:

```go
// ErrConnectionPairExists signals a 409 from create: a connection for this
// source/destination pair already exists, so the externalId claim was refused.
var ErrConnectionPairExists = fmt.Errorf("connection already exists for this source and destination pair")
```

mapped in `Create` the same way this branch maps the set endpoint.

sivashanmukh (commented):
Reviewed as part of the full stack (#738 → #739 → #742 → #748 → #750 → #753 → #754). Verified at the stack tip: `go build ./...` clean, `go vet` clean on all touched packages, and `go test` green for every changed package.

This PR is mechanically faithful to the existing shapes — the `list` + functional-options + `Paging{Next: path}` construction matches `accounts.list` exactly, `SetExternalID` matches `destinations.SetExternalID`, and `fmt.Errorf` for the sentinel matches `ErrResourceNotFound` in this package. No style findings.

The findings below are all semantic, and they come from the backend change this client talks to (rudderlabs/rudder-config-backend#6576), which I reviewed in parallel.

**Two out-of-diff drive-bys**, neither introduced here, both worth a cleanup ticket:

- `cli/internal/project/specs/spec.go:128` — `ParseSpecReference` opens with `fmt.Printf("ref: %s\n kindMap: %v\n", ref, kindMap)`, printing debug output to stdout from a library function on every reference parse. Same function has an `unkown kind` typo at line 138.
- `CLAUDE.md` still contains an unresolved merge conflict marker (`>>>>>>> 03ec8f27`) around line 60, with a duplicated code block above it.

api/client/connections.go
```diff
@@ -79,3 +124,21 @@ func (s *connections) Update(ctx context.Context, connection *Connection) (*Conn
 func (s *connections) Delete(ctx context.Context, id string) error {
 	return s.service.delete(ctx, id)
 }
+
+func (s *connections) SetExternalID(ctx context.Context, id, externalID string) error {
```
sivashanmukh, line 145: Not blocking, and out of scope for this PR — recording it while it is visible.

`connections` is now the only one of the three externalId setters in this package that maps the conflict to a sentinel. `accounts.SetExternalID` and `destinations.SetExternalID` both return a wrapped generic error, so their import flows (`providers/accounts/handler.go:220`, `providers/destination/handler.go:365`) cannot branch on a conflict either — they surface it as a generic import failure.

Once the shape here settles, it is worth lifting to the other two rather than leaving three externalId setters with three different error contracts.

sivashanmukh (approved):

api/client/connections.go
dhawal1248: Done in 26b48dd1 — took your wording as-is.

The same commit adds a sibling `ErrConnectionPairExists` for the create-side 409, documented the same way. Detail in the reply on the 409 branch below.

api/client/connections.go
dhawal1248: Both parts are in 26b48dd1.

**1.** `SetExternalID` now returns `fmt.Errorf("%w: %w", ErrExternalIDAlreadyInUse, err)`, so `errors.Is` keeps working for existing callers and the backend's message survives for anyone who needs to tell the two conflict conditions apart.

**2.** Added `ErrConnectionPairExists` and mapped the create 409 the same way.

One tradeoff worth flagging rather than burying: `Create` currently maps **every** 409 to `ErrConnectionPairExists`, not just pair conflicts. That mirrors `SetExternalID`, which already maps every 409 despite covering two conditions, and the wrapped error preserves the backend's message either way. But if the backend ever returns a 409 from create for an unrelated reason, the sentinel would be misleading. Happy to narrow it to a message match — I left it broad for consistency with the set endpoint.

On the machine-readable `code` you asked the backend for: `APIError` already parses it (`ErrorCode string` with a `json:"code"` tag, `api/client/common.go:28`), so when those codes land, branching on them is a field access. No client plumbing needed.

The adoption behaviour in #754 is deliberately not in this PR — being scoped separately.

api/client/connections.go
dhawal1248: Agreed, and not done here — you called it out of scope, and I'd rather not widen this PR.

Recording the shape while it's fresh: `accounts.SetExternalID` and `destinations.SetExternalID` both wrap into a generic error, so `providers/accounts/handler.go` and `providers/destination/handler.go` can't branch on a conflict during import. Lifting the sentinel contract from here to both is the fix — ideally once the create-side sentinel added in this PR has settled, so all three land on the same shape rather than converging twice.

PR #739, opened 2026-08-17
feat(event-stream): define event-stream-connections spec kind

dhawal1248:
## 🔗 Ticket

Resolves [DEX-649](https://linear.app/rudderstack/issue/DEX-649/3-define-the-event-stream-connections-spec-kind)

---

## Summary

Defines the new `event-stream-connections` spec kind: a list of pure links between event stream sources and destinations, handled by a new `connection` handler inside the existing event-stream provider. Parsed `source`/`destination` references become `PropertyRef`s that resolve to the referenced resource's remote ID from state, so the resource graph automatically derives dependency edges (connection depends on both endpoints), correct create/delete ordering, and cycle detection. Stacked on top of #738.

---

## Changes

- New `cli/internal/providers/event-stream/connection` handler package mirroring the `source` handler style: spec types (`id`, `source`, `destination`, optional `enabled` defaulting to `true`), `ParseSpec` emitting one URN per list entry, and strict `ErrorUnused` decoding so a `config` key (or any unknown field) fails at parse time — connections are pure links; settings live on the destination spec
- Source refs (`#event-stream-source:<id>`) resolve via the source's legacy state output map; destination refs (`#destination:<id>`) resolve via the destination provider's typed state (`handler.CreatePropertyRef`, mirroring the transformation ref pattern)
- Connections are plain data-map resources; the graph derives dependency edges from the two endpoint refs, and the syncer's map-based dereference resolves both to remote IDs before lifecycle calls
- Registers the `event-stream-connections` kind (v1-only, no legacy versions) in the event-stream provider behind a new `connectionSupport` experimental flag — with the flag off the kind is not a supported spec; lifecycle methods return a clear not-implemented error pending DEX-650, and remote-load methods return empty results so apply/import keep working for other event stream resource types
- Teaches the map-based `state.Dereference` to honor a ref's `Resolve` func (the same logic `DereferenceByReflection` already had) so legacy-provider resources can reference typed-state (modern framework) resources — this is what lets the `#destination:` ref resolve without special routing
- Adds the new kind to the project-level rule doc fragments (`duplicate-urn`, `metadata-syntax-valid`, `manifest-inline-conflict`) required by the rule-catalog drift gates

---

## Testing

- Unit tests: handler tests cover the ticket's YAML example, enabled default, empty-list validity, missing `connections`/`id`/`source`/`destination`, `config` and unknown-key rejection, malformed/wrong-kind refs, destination ref resolution, graph dependency edges + cycle detection, and end-to-end ref resolution through state (legacy source output map + typed destination state); provider tests updated for the new kind/type/pattern expectations
- `make lint` (0 issues) and `make test` (all unit packages pass); connection package statement coverage 98.8%

---

## Risk / Impact

Low
Notes: The new kind is off by default (behind the `connectionSupport` experimental flag); with it on, apply lifecycle intentionally errors with a clear message until DEX-650 lands. Validation rules for the kind follow in DEX-652/653 (validate tags are already in place), import/export in DEX-654.

---

## Checklist

- [x] Ticket linked
- [x] Tests added/updated

### DIFF
--- a/cli/internal/providers/event-stream/connection/handler.go
+++ b/cli/internal/providers/event-stream/connection/handler.go
@@ -54,7 +54,7 @@ func (h *Handler) ParseSpec(_ string, s *specs.Spec) (*specs.ParsedSpec, error) 
 			return nil, fmt.Errorf("id not found in connection at index %d", i)
 		}
 		entries = append(entries, specs.URNEntry{
-			URN:             resources.URN(id, ResourceType),
+			URN:             resources.URN(id, EventStreamConnectionResourceType),
 			JSONPointerPath: fmt.Sprintf("/spec/connections/%d/id", i),
 		})
 	}
@@ -63,9 +63,8 @@ func (h *Handler) ParseSpec(_ string, s *specs.Spec) (*specs.ParsedSpec, error) 
 
 func (h *Handler) LoadSpec(_ string, s *specs.Spec) error {
 	spec := &ConnectionsSpec{}
-	// Use strict decoding to reject unknown fields — a config key on a
-	// connection must fail at parse time: connections are pure links and
-	// connection settings live on the destination spec.
+
+	// Use strict decoding to reject unknown fields.
 	decoderConfig := &mapstructure.DecoderConfig{
 		ErrorUnused: true,
 		Result:      spec,
@@ -77,23 +76,8 @@ func (h *Handler) LoadSpec(_ string, s *specs.Spec) error {
 	if err := decoder.Decode(s.Spec); err != nil {
 		return fmt.Errorf("decoding event stream connections spec: %w", err)
 	}
-	// Import support lands with DEX-654: reject inline import metadata now so
-	// it is never silently treated as a plain create.
-	metadata, err := s.CommonMetadata()
-	if err != nil {
-		return fmt.Errorf("getting common metadata: %w", err)
-	}
-	if metadata.Import != nil {
-		return fmt.Errorf("import metadata is not supported for %s yet", ResourceKind)
-	}
-	// A nil slice means the connections key was absent: fail loudly instead of
-	// letting the file silently contribute nothing. An explicit empty list
-	// stays valid so removing the last connection does not invalidate the spec.
-	if spec.Connections == nil {
-		return fmt.Errorf("connections not found in event stream connections spec")
-	}
-	for i := range spec.Connections {
-		resource, err := h.loadConnection(&spec.Connections[i], i)
+	for _, c := range spec.Connections {
+		resource, err := h.loadConnection(c)
 		if err != nil {
 			return err
 		}
@@ -104,17 +88,10 @@ func (h *Handler) LoadSpec(_ string, s *specs.Spec) error {
 	return nil
 }
 
-func (h *Handler) loadConnection(c *ConnectionSpec, index int) (*connectionResource, error) {
-	if c.LocalID == "" {
-		return nil, fmt.Errorf("id is required for connection at index %d", index)
-	}
-	if c.Source == "" {
-		return nil, fmt.Errorf("source is required for connection %q", c.LocalID)
-	}
-	if c.Destination == "" {
-		return nil, fmt.Errorf("destination is required for connection %q", c.LocalID)
-	}
-
+// loadConnection builds the graph-side resource for one entry. Required-field
+// checks are owned by the validation rules driven by the spec's validate tags
+// (DEX-652); only reference parsing can fail here.
+func (h *Handler) loadConnection(c ConnectionSpec) (*connectionResource, error) {
 	sourceRef, err := parseSourceRef(c.Source)
 	if err != nil {
 		return nil, fmt.Errorf("connection %q: parsing source reference: %w", c.LocalID, err)
@@ -152,10 +129,10 @@ func (h *Handler) GetResources() ([]*resources.Resource, error) {
 	for _, c := range h.resources {
 		r := resources.NewResource(
 			c.LocalID,
-			ResourceType,
+			EventStreamConnectionResourceType,
 			resources.ResourceData{},
 			[]string{},
-			resources.WithResourceFileMetadata(fmt.Sprintf("#%s:%s", ResourceKind, c.LocalID)),
+			resources.WithResourceFileMetadata(fmt.Sprintf("#%s:%s", EventStreamConnectionResourceKind, c.LocalID)),
 			resources.WithRawData(c),
 		)
 		result = append(result, r)
@@ -163,9 +140,7 @@ func (h *Handler) GetResources() ([]*resources.Resource, error) {
 	return result, nil
 }
 
-// parseSourceRef parses a scalar "#event-stream-source:<id>" reference. The
-// ref resolves via the source's legacy state output map, which carries the
-// remote id under "id".
+// parseSourceRef parses a scalar "#event-stream-source:<id>" reference.
 func parseSourceRef(ref string) (*resources.PropertyRef, error) {
 	id, err := refID(ref, source.ResourceKind)
 	if err != nil {
@@ -213,22 +188,9 @@ func refID(ref string, kind string) (string, error) {
 	return parts[1], nil
 }
 
-// LoadImportMetadata rejects central import-manifest entries that target this
-// kind — import support lands with DEX-654. Entries for other kinds pass
-// through untouched: the provider broadcasts the full manifest to every
-// handler.
-func (h *Handler) LoadImportMetadata(m *specs.WorkspacesImportMetadata) error {
-	if m == nil {
-		return nil
-	}
-	prefix := ResourceType + ":"
-	for _, workspace := range m.Workspaces {
-		for _, resource := range workspace.Resources {
-			if strings.HasPrefix(resource.URN, prefix) {
-				return fmt.Errorf("import is not supported for %s yet (manifest entry %s)", ResourceKind, resource.URN)
-			}
-		}
-	}
+// LoadImportMetadata is a no-op: import support for connections lands with
+// DEX-654.
+func (h *Handler) LoadImportMetadata(_ *specs.WorkspacesImportMetadata) error {
 	return nil
 }
 
--- a/cli/internal/providers/event-stream/connection/handler.go
+++ b/cli/internal/providers/event-stream/connection/handler.go
@@ -21,9 +21,7 @@ import (
 )
 
 // ErrNotImplemented guards the lifecycle surface until DEX-650 wires the
-// connections API client; spec parsing and graph construction work today. The
-// event-stream provider returns it from the raw lifecycle methods the syncer
-// routes RawData resources through.
+// connections API client; spec parsing and graph construction work today.
 var ErrNotImplemented = errors.New("event stream connection apply support is not implemented yet")
 
 type Handler struct {
@@ -121,19 +119,23 @@ func (h *Handler) MigrateSpec(s *specs.Spec) (*specs.Spec, error) {
 	return s, nil
 }
 
-// GetResources attaches each connection as raw data so the syncer's
-// reflection-based dereference resolves the typed destination ref and the
-// graph derives dependency edges on both endpoints.
+// GetResources emits each connection as a plain data-map resource; the graph
+// derives dependency edges from the two endpoint refs in the map, and the
+// syncer dereferences them to remote ids before the lifecycle calls.
 func (h *Handler) GetResources() ([]*resources.Resource, error) {
 	result := make([]*resources.Resource, 0, len(h.resources))
 	for _, c := range h.resources {
+		data := resources.ResourceData{
+			SourceKey:      c.Source,
+			DestinationKey: c.Destination,
+			EnabledKey:     c.Enabled,
+		}
 		r := resources.NewResource(
 			c.LocalID,
 			EventStreamConnectionResourceType,
-			resources.ResourceData{},
+			data,
 			[]string{},
 			resources.WithResourceFileMetadata(fmt.Sprintf("#%s:%s", EventStreamConnectionResourceKind, c.LocalID)),
-			resources.WithRawData(c),
 		)
 		result = append(result, r)
 	}
--- a/cli/internal/providers/event-stream/connection/handler.go
+++ b/cli/internal/providers/event-stream/connection/handler.go
@@ -20,9 +20,11 @@ import (
 	"github.com/rudderlabs/rudder-iac/cli/internal/resources/state"
 )
 
-// errNotImplemented guards the lifecycle surface until DEX-650 wires the
-// connections API client; spec parsing and graph construction work today.
-var errNotImplemented = errors.New("event stream connection apply support is not implemented yet")
+// ErrNotImplemented guards the lifecycle surface until DEX-650 wires the
+// connections API client; spec parsing and graph construction work today. The
+// event-stream provider returns it from the raw lifecycle methods the syncer
+// routes RawData resources through.
+var ErrNotImplemented = errors.New("event stream connection apply support is not implemented yet")
 
 type Handler struct {
 	resources map[string]*connectionResource
@@ -74,6 +76,15 @@ func (h *Handler) LoadSpec(_ string, s *specs.Spec) error {
 	}
 	if err := decoder.Decode(s.Spec); err != nil {
 		return fmt.Errorf("decoding event stream connections spec: %w", err)
+	}
+	// Import support lands with DEX-654: reject inline import metadata now so
+	// it is never silently treated as a plain create.
+	metadata, err := s.CommonMetadata()
+	if err != nil {
+		return fmt.Errorf("getting common metadata: %w", err)
+	}
+	if metadata.Import != nil {
+		return fmt.Errorf("import metadata is not supported for %s yet", ResourceKind)
 	}
 	// A nil slice means the connections key was absent: fail loudly instead of
 	// letting the file silently contribute nothing. An explicit empty list
@@ -144,7 +155,7 @@ func (h *Handler) GetResources() ([]*resources.Resource, error) {
 			ResourceType,
 			resources.ResourceData{},
 			[]string{},
-			resources.WithResourceFileMetadata(fmt.Sprintf("#%s:%s", ResourceType, c.LocalID)),
+			resources.WithResourceFileMetadata(fmt.Sprintf("#%s:%s", ResourceKind, c.LocalID)),
 			resources.WithRawData(c),
 		)
 		result = append(result, r)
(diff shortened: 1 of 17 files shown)


cli/internal/providers/event-stream/connection/model.go
```diff
@@ -0,0 +1,39 @@
+package connection
+
+import (
+	"github.com/rudderlabs/rudder-iac/cli/internal/resources"
+)
+
+const (
+	ResourceType = "event-stream-connection"
```
dhawal1248: rename this to `EventStreamConnectionResourceType` , same for the kind

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	// Import support lands with DEX-654: reject inline import metadata now so
+	// it is never silently treated as a plain create.
+	metadata, err := s.CommonMetadata()
+	if err != nil {
+		return fmt.Errorf("getting common metadata: %w", err)
+	}
+	if metadata.Import != nil {
+		return fmt.Errorf("import metadata is not supported for %s yet", ResourceKind)
+	}
```
dhawal1248: remove these checks for import metadata, instead check the experimental flag for ConnectionsSupport and make the decisions whether it is a supported spec or not based on that

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+	// A nil slice means the connections key was absent: fail loudly instead of
+	// letting the file silently contribute nothing. An explicit empty list
+	// stays valid so removing the last connection does not invalidate the spec.
+	if spec.Connections == nil {
+		return fmt.Errorf("connections not found in event stream connections spec")
+	}
```
dhawal1248: can this be acheived via the validation tags?

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+	for i := range spec.Connections {
+		resource, err := h.loadConnection(&spec.Connections[i], i)
```
dhawal1248: is it okay to send the address of a pointer variable to a function call like this?

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+func (h *Handler) loadConnection(c *ConnectionSpec, index int) (*connectionResource, error) {
+	if c.LocalID == "" {
+		return nil, fmt.Errorf("id is required for connection at index %d", index)
+	}
+	if c.Source == "" {
+		return nil, fmt.Errorf("source is required for connection %q", c.LocalID)
+	}
+	if c.Destination == "" {
+		return nil, fmt.Errorf("destination is required for connection %q", c.LocalID)
+	}
```
dhawal1248: will these checks not be done by the validation framework via the validator tags? check what do we do for such cases in other handlers/providers

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+		result = append(result, r)
+	}
+	return result, nil
+}
+
+// parseSourceRef parses a scalar "#event-stream-source:<id>" reference. The
+// ref resolves via the source's legacy state output map, which carries the
+// remote id under "id".
+func parseSourceRef(ref string) (*resources.PropertyRef, error) {
+	id, err := refID(ref, source.ResourceKind)
```
dhawal1248, line 148: do we need to add this new function, dont we have common functions to work with ref strings? how do other providers do this?

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,276 @@
...
+
+// parseDestinationRef parses a scalar "#destination:<id>" reference into a
+// PropertyRef whose Resolve function reads DestinationState.ID, mirroring the
+// destination provider's transformation ref.
+func parseDestinationRef(ref string) (*resources.PropertyRef, error) {
+	id, err := refID(ref, destination.DestinationSpecKind)
+	if err != nil {
+		return nil, err
+	}
+	propertyRef := handler.CreatePropertyRef(
```
dhawal1248, line 166: lets use the simple property ref creation method that we used in parseSourceRef

cli/internal/providers/event-stream/provider.go
```diff
@@ -219,6 +225,38 @@ func (p *Provider) Delete(ctx context.Context, ID string, resourceType string, s
 	return handler.Delete(ctx, ID, state)
 }
 
+// Raw lifecycle: the syncer routes resources carrying RawData through these —
```
dhawal1248: lets use the same Create functions as the event stream source provider - normal Create, Update functions..we will migrate all old providers to new later

cli/internal/providers/event-stream/connection/model.go
dhawal1248: Renamed to `EventStreamConnectionResourceType` / `EventStreamConnectionResourceKind` in c17e0824; all references updated.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Removed both import-metadata checks. Added a `connectionSupport` experimental flag (`ConnectionSupport` in `ExperimentalConfig` — matching the DEX-650 ticket's spelling and the `destinationSupport`/`accountSupport` naming convention). `eventstream.New` now takes the flag and only registers the connections kind/handler when it's on; with it off these specs fail as `unsupported kind`. The gen-rule-docs CI workflow and the app catalog test enable `RUDDERSTACK_X_CONNECTION_SUPPORT` so the rule-doc fragments stay covered.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Removed the runtime check; `Connections` now carries `validate:"required"` alongside the required tags on `id`/`source`/`destination`, for the DEX-652 syntactic rule to enforce — same split as SourceSpec/DestinationSpec, whose required-ness lives in their spec-syntax rules rather than LoadSpec. `ParseSpec` still errors on a missing `connections` key since URN extraction has nothing to walk.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: It was the address of a slice element (`&spec.Connections[i]`), not of a pointer variable — legal Go, but unnecessary here. `loadConnection` now takes `ConnectionSpec` by value.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Checked the other handlers: required-ness is enforced by the spec-syntax rules via the validate tags (`rules/source/source_spec_valid.go`, `destination/rule_spec_syntax_valid.go`), not in LoadSpec. Dropped the guards — the validate tags are in place for DEX-652's rule. Empty `source`/`destination` still fail at parse time, as unparseable references.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: No shared ref-string helper exists today — each provider parses its own: `destination/handler.go` `parseTransformationRef`, `datagraph/provider.go` `parseModelReference`, datacatalog via regexes in `rules/constants.go`. `refID` mirrors the destination one and stays package-local like the others.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: A plain `PropertyRef{Property: "id"}` can't resolve a destination. The destination provider is on the typed-handler framework, so its state entries carry only `OutputRaw *DestinationState`; the `Data()` map both dereference paths use for property lookup is empty (`resources/state/state.go` `resolvePropertyRef` / `dereferenceValue`, `provider/handler/basehandler.go` `MapRemoteToState`). The tracking-plan ref in the source handler gets away with the simple form only because tracking plans are a legacy provider whose state output map carries `id`. `handler.CreatePropertyRef` is the existing shared helper for referencing typed-state resources — destination uses exactly this shape for its transformation ref (`createTransformationRef`, `destination/handler.go`). `TestConnectionRefResolution` proves both endpoint refs resolve from state; with the simple form it fails with `property 'id' does not exist in resource 'destination:s3'`. Keeping the typed ref for now — it collapses to the simple form once providers migrate to the modern framework.

cli/internal/providers/event-stream/provider.go
dhawal1248: The handler already has the source-style surface — normal `Create`/`Update`/`Delete`/`List`/`Import` methods registered in the same `handlers` map — and DEX-650 implements those functions, like the source handler's. The provider-level `*Raw` overrides exist because the syncer routes any resource carrying RawData through `CreateRaw`/`UpdateRaw`/`DeleteRaw`/`ImportRaw` (`syncer.go` `createOperation`), and RawData is what makes the typed destination ref resolvable (see the ref thread above); without the overrides those calls dead-end in `EmptyProvider`'s generic `not implemented`. In DEX-650 the raw methods stay thin: extract the resolved data, call the handler's normal functions. They disappear entirely when the old providers migrate to the new framework.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Update after the follow-up discussion: connections are now plain data-map resources (df0918e6) — the map-based `state.Dereference` learned the same `Resolve` handling `DereferenceByReflection` already had, so the typed destination ref resolves on the normal path. The ref itself still carries the Resolve func (that part is unavoidable while destination state is typed), but all the RawData routing is gone.

cli/internal/providers/event-stream/provider.go
dhawal1248: Done in df0918e6 — after the dereference fix, connections travel the normal lifecycle path: no RawData, the provider `*Raw` overrides are deleted, and `provider.Create/Update/Delete/Import` route straight to the handler's normal source-style methods, which DEX-650 implements with already-resolved remote IDs in the data map.

cli/internal/providers/event-stream/provider.go
```diff
@@ -54,14 +55,20 @@ type Provider struct {
 	handlers   map[string]handler
 }
 
-func New(client esClient.EventStreamStore) *Provider {
+func New(client esClient.EventStreamStore, connectionSupport bool) *Provider {
```
sivashanmukh: **Third mechanism for gating an experimental flag.**

The repo already has two:

- gate at the registration site in `dependencies.go` — `if cfg.ExperimentalFlags.DestinationSupport { … }` (line 243), `AccountSupport` (line 255)
- read the flag internally where the sub-feature lives — `providers/datacatalog/provider.go:363` does exactly this for `EventRuleIncludes`, which is the closest analogue to gating one kind inside a provider

A positional bool is neither, and it makes the call site unreadable: `esProvider.New(client, true)` says nothing about what is being enabled. It also does not extend — the next flag makes it `New(client, true, false)`.

A functional option matches the style this stack already uses in #738 (`WithConnectionsHasExternalID`) and elsewhere in the repo (`client.WithHTTPClient`, `catalog.WithConcurrency`):

```go
type Option func(*Provider)

// WithConnectionSupport registers the event-stream-connections kind. Gated
// behind the connectionSupport experimental flag.
func WithConnectionSupport() Option {
	return func(p *Provider) {
		p.kindToType[connectionHandler.EventStreamConnectionResourceKind] = connectionHandler.EventStreamConnectionResourceType
		p.handlers[connectionHandler.EventStreamConnectionResourceType] = connectionHandler.NewHandler()
	}
}

func New(client esClient.EventStreamStore, opts ...Option) *Provider {
	p := &Provider{ /* … */ }
	p.handlers[sourceHandler.ResourceType] = sourceHandler.NewHandler(client, importDir)
	for _, opt := range opts {
		opt(p)
	}
	return p
}
```

with `dependencies.go` reading:

```go
var esOpts []esProvider.Option
if cfg.ExperimentalFlags.ConnectionSupport {
	esOpts = append(esOpts, esProvider.WithConnectionSupport())
}
esp := esProvider.New(esClient.NewRudderEventStreamStore(c), esOpts...)
```

If you would rather not add an option type, the `datacatalog` precedent (read `config.GetConfig().ExperimentalFlags.ConnectionSupport` inside `New`) is also fine. Either beats the bool — the ask is just that it not be a third pattern.

cli/internal/providers/event-stream/provider.go
```diff
@@ -90,7 +97,10 @@ func (p *Provider) SupportedKinds() []string {
 func (p *Provider) SupportedMatchPatterns() []rules.MatchPattern {
 	var patterns []rules.MatchPattern
 	for kind := range p.kindToType {
-		patterns = append(patterns, prules.LegacyVersionPatterns(kind)...)
+		// event-stream-connections is a new kind with no legacy spec versions
+		if kind != connectionHandler.EventStreamConnectionResourceKind {
+			patterns = append(patterns, prules.LegacyVersionPatterns(kind)...)
```
sivashanmukh: **This branch grows with every v1-only kind.**

Special-casing by kind identity inside the loop means the next kind without legacy versions adds another `&&`. A property of the kind is cheaper to express as data:

```go
// kinds that were introduced after legacy spec versions were retired
var kindsWithoutLegacyVersions = map[string]struct{}{
	connectionHandler.EventStreamConnectionResourceKind: {},
}

func (p *Provider) SupportedMatchPatterns() []rules.MatchPattern {
	var patterns []rules.MatchPattern
	for kind := range p.kindToType {
		if _, v1Only := kindsWithoutLegacyVersions[kind]; !v1Only {
			patterns = append(patterns, prules.LegacyVersionPatterns(kind)...)
		}
		patterns = append(patterns, prules.V1VersionPatterns(kind)...)
	}
	return patterns
}
```

Minor — flagging it because the second kind is where the pattern gets set.

sivashanmukh (commented):
Reviewed as part of the full stack. Verified at the stack tip: build, vet and tests all clean.

The `state.Dereference` change is the part I looked at hardest, since it is shared infrastructure, and it is correct: it mirrors `resolvePropertyRef` in `reflection.go` faithfully, and it is backward compatible because legacy refs carry `Resolve == nil` and keep the old map-lookup path. Right change, right place.

Findings inline: one deviation (flag threading), two duplications, one minor scaling nit.

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,240 @@
...
+		func(state *destination.DestinationState) (string, error) {
+			if state.ID == "" {
+				return "", fmt.Errorf("destination state has empty ID")
+			}
+			return state.ID, nil
+		},
+	)
+	// Stamp the "id" property so the differ's comparePropertyRefs sees a
+	// stable shape on both the spec and state sides.
+	propertyRef.Property = "id"
```
sivashanmukh, line 184: **Copied verbatim from `destination/handler.go:661-675`, comment included.**

`createTransformationRef` there does the same `handler.CreatePropertyRef(...)` then `ref.Property = "id"` with the identical justification about `comparePropertyRefs` seeing a stable shape. This is now the second occurrence of the same workaround, which is usually the point at which it belongs in the helper rather than in each caller:

```go
// in cli/internal/provider/handler/basehandler.go
func CreatePropertyRef[State any](
	urn string,
	property string,
	extractor func(*State) (string, error),
) *resources.PropertyRef {
	return &resources.PropertyRef{
		URN:      urn,
		Property: property,
		Resolve:  /* … unchanged … */,
	}
}
```

That is a signature change across six call sites (`datagraph` ×4, `destination`, `testutils/example`), so it is fair to say it does not belong in this PR. But please either do it here or leave a TODO pointing at the duplication — the third copy is the one that gets it wrong.

I verified the workaround is genuinely needed, for what it is worth: `differ.comparePropertyRefs` (`syncer/differ/diff.go:330`) explicitly excludes the `Resolve` field, so spec-side and state-side refs compare on URN + Property alone. The comment is accurate.

cli/internal/providers/event-stream/connection/handler.go
```diff
@@ -0,0 +1,240 @@
...
+	return propertyRef, nil
+}
+
+// refID extracts <id> from a scalar "#<kind>:<id>" reference.
+func refID(ref string, kind string) (string, error) {
```
sivashanmukh, line 196: **Fourth implementation of `#kind:id` parsing in the repo, and the second in this stack.**

Existing:
- `destination/handler.go:641-658` — `parseTransformationRef`, hardcoded to one kind
- `project/specs/spec.go:129` — `ParseSpecReference`, exported and general, takes a `kindMap`

Added by this stack:
- this `refID`
- `scalarRefRegex` in #750 (`rules/connection/connection_spec_valid.go:21`)

The two added here are not equivalent, which is the part that concerns me. `refID` uses `SplitN` and accepts any kind charset and a multi-line id; `scalarRefRegex` is `^#([a-zA-Z0-9_-]+):(.+)$`, which constrains the kind and rejects a multi-line id (`.` does not match `\n`). So a ref like `#event-stream-source:a\nb` is rejected by the validation rule and accepted by this loader.

That is harmless today because validate runs before load, but two parsers for one grammar is exactly how validate-time and apply-time behaviour drift apart. Since `refID` is already generic in the kind, the cheapest fix is to export it and have #750's rule call it:

```go
// ParseRef extracts <id> from a scalar "#<kind>:<id>" reference.
func ParseRef(ref, kind string) (string, error) { /* current body */ }
```

then in #750, replace `scalarRefRegex`-based matching with `ParseRef` plus a separate "which kind did we actually get" probe. Longer term `specs.ParseSpecReference` is the natural single home for all four, once someone untangles its `kindMap` requirement.

sivashanmukh (approved):

cli/internal/providers/event-stream/provider.go
dhawal1248: Done in 884e8fb3 — went with the functional option.

Picked it over the `datacatalog` "read the flag inside `New`" alternative for one concrete reason: `provider_test.go` has an explicit connection-support-disabled case, and reading global config inside `New` would force that test to mutate global state to exercise it. The option keeps the disabled path expressible without touching globals.

23 call sites in total — one in `dependencies.go`, 21 in `provider_test.go`, and one in `ruledoc_test.go` that was easy to miss and only surfaced under `go vet`.

cli/internal/providers/event-stream/provider.go
dhawal1248: Done in 884e8fb3, as suggested — `kindsWithoutLegacyVersions` as a package-level map, and the loop does a lookup instead of comparing kind identity.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Took the TODO option, in 884e8fb3.

The signature change reaches 7 call sites (`datagraph` ×4, `destination`, this one, and `testutils/example`), and this stack is 7 PRs deep with the lower ones still moving — threading it through now means rebasing the whole stack for a change that isn't load-bearing for the feature. The TODO sits on `parseDestinationRef`, names `destination`'s `createTransformationRef` as the other copy, and states the fix: take `property` as a `CreatePropertyRef` parameter so callers can't forget it.

Thanks for verifying the workaround is genuinely needed against `comparePropertyRefs` — that saved a round trip.

cli/internal/providers/event-stream/connection/handler.go
dhawal1248: Done, across this PR and #750 — your drift analysis was right, and it landed close to the cheap direction you offered.

Here (9342eab3): `refID` no longer hand-rolls `TrimPrefix`/`SplitN`. It matches on `^#([a-zA-Z0-9_-]+):(.+)$`, the same pattern the rule uses, so the loader is no longer more permissive than the validator. `#event-stream-source:a\nb` is now rejected on both sides.

In #750 (5a08bfcd) that pattern is exported as `esConnection.ScalarRefRegex` and the rule consumes it, so there's one definition rather than two that happen to agree.

That's slightly different from your suggestion of exporting `refID` as `ParseRef`: the rule's wrong-kind diagnostic needs the captured kind to say "got a '%s' reference", and a `ParseRef` returning only the id would have needed a second probe to recover it. Exporting the pattern kept both behaviours against one source of truth. Say the word if you'd rather it were the function.

`destination/handler.go:641` and `specs.ParseSpecReference` are untouched and still separate implementations — out of scope here, but the consolidation target is clearer now.

